PR #6, opened 2022-06-16
Support fragment caching

joeldrapper:
Very early work-in-progress support for fragment caching in Phlex.


lib/phlex/component.rb
```diff
@@ -1,27 +1,112 @@
 # frozen_string_literal: true
+require "digest"
+require "byebug"
 
 module Phlex
   class Component
     include Node, Context
 
+    class << self
```
joeldrapper, line 9: All this stuff should be in a separate module.

lib/phlex/component.rb
```diff
@@ -1,27 +1,112 @@
...
+
+      private
+
+      def dependant_ancestor_checksums
+        dependant_ancestors.map do |ancestor|
+          Digest::MD5.hexdigest(
+            File.read(
+              Module.const_source_location(ancestor.name).first
+            )
+          )
```
joeldrapper, line 38: We could make a local checksum cache so we only ever need to build a checksum once per file path.

lib/phlex/component.rb
```diff
@@ -1,27 +1,112 @@
...
+      end
+
+      def dependant_ancestors
+        ancestors.map
+          .reject { [self, Object, BasicObject, Kernel].include? _1 }
+          .reject { _1.name.start_with? "Phlex::" }
+      end
+
+      def direct_subcomponents
+        file.scan(/component\s([A-Z]+[A-Za-z:]*)/).flatten.map { const_get(_1) }
```
joeldrapper, line 49: `const_get` here won't look up relative constants based on the namespace, so we need to write something else that will do that.

lib/phlex/component.rb
```diff
@@ -1,27 +1,112 @@
...
+        Rails.cache.fetch(cache_key) { super }
+      end
+
+      def cache_key
+        if @_content
+          content = Digest::MD5.hexdigest(
+            File.read(
+              @_content&.source_location.first
+            )
+          )
```
joeldrapper, line 85: We need to get the sub components that might have been called inside the content block too.

joeldrapper:
I’m merging this as its without tests because Phlex is still being sketched out in early beta.

PR #15, opened 2022-07-05
Performance specs

joeldrapper:


lib/phlex/tag/class_collector.rb
```diff
@@ -6,9 +6,9 @@ def initialize(context, tag)
         @tag = tag
       end
 
-      def method_missing(name, content = nil, *args, **attributes, &block)
+      def method_missing(name, content = nil, **attributes, &block)
```
joeldrapper, line 9: This created a useless Array every time we chained a class.

PR #27, opened 2022-07-15
Use Zeitwerk to load constants

alexandreruban:
References #25

@joeldrapper I didn't see you self assigned the Zeitwerk issue. Here is a draft. I can close this PR if you are working on it, or otherwise let me know if you're up for a peer programming! 😄


joeldrapper:
This is great! I hadn't started yet. 😅

I was thinking of updating `Tag` to take the tag name as an argument and then get rid of all of these subclasses, so that might make this easier.

lib/phlex/tag/standard_element.rb
```diff
@@ -3,8 +3,105 @@ class Tag
...
+        small
+        span
+        strong
+        sub
+        summary
+        sup
+        table
+        tbody
+        td
+        template
```
joeldrapper, line 89: This one is interesting, you can only use it with `super`. 🤣

lib/phlex/tag.rb
```diff
@@ -7,13 +7,10 @@ class Tag
...
 
-    class << self
-      def tag_name
-        @tag_name ||= name.split(NAMESPACE_DELINEATOR).last.downcase
-      end
-    end
+    attr_reader :name
 
-    def initialize(**attributes)
+    def initialize(name:, **attributes)
```
joeldrapper, line 12: Let's use a positional argument for the name so the `name` keyword argument can be a tag attribute, e.g. `<div name="something">`.

alexandreruban: Nice catch! Just added the change in the last commit.

joeldrapper (approved):

PR #32, opened 2022-07-18
Add configuration for underscore to dash conversion in class names

saksham-jain:
- Add configuration class having convert_underscores_to_dashes instance variable
- Add configure method which accepts block to set configurations
- Check convert_underscores_to_dashes config before converting underscores to dashes
- Add specs to check the configuration


lib/phlex/tag.rb
```diff
@@ -53,7 +53,7 @@ def attributes
 
     def classes
       return if @classes.empty?
-      @classes.gsub!(UNDERSCORE, DASH)
+      @classes.gsub!(UNDERSCORE, DASH) if Phlex.configuration.convert_underscores_to_dashes.eql?(true)
```
joeldrapper, line 56:
```suggestion
      @classes.gsub!(UNDERSCORE, DASH) if Phlex.configuration.convert_underscores_to_dashes
```

spec/phlex/component_spec.rb
```diff
@@ -178,6 +178,26 @@ def template
...
+      Class.new Phlex::Component do
+        def template
+          div class: "a_b"
+        end
+      end
+    end
+
+    it "when convert_underscores_to_dashes config is true convert underscores to dashes and produces the classes" do
+      expect(output).to eq %{<div class="a-b"></div>}
+    end
```
joeldrapper, line 192: Can we add and extra `describe` block to these examples so it doesn't read "it when"?

```suggestion
    describe "when configured to convert_underscores_to_dashes"
      it "converts underscores to dashes and produces the classes" do
        expect(output).to eq %{<div class="a-b"></div>}
      end
    end
```

spec/phlex/component_spec.rb
```diff
@@ -178,6 +178,26 @@ def template
...
+
+    it "when convert_underscores_to_dashes config is false do not convert underscores to dashes and produces the classes" do
+      Phlex.configure { |config| config.convert_underscores_to_dashes = false }
```
joeldrapper, line 195: We should be able to put the configuration in an around each hook rather than include it in the `it`. https://relishapp.com/rspec/rspec-core/v/2-0/docs/hooks/around-hooks

lib/phlex.rb
```diff
@@ -2,6 +2,7 @@
 
 require "active_support/core_ext/string/output_safety"
 require "zeitwerk"
+require "phlex/configuration"
```
joeldrapper, line 5: You shouldn't need this require as Zeitwerk will autoload it.

joeldrapper (commented):
Hey @saksham-jain this is looking really good. Thank you for working on this! ❤️ 

I’ve left a few comments.

spec/phlex/component_spec.rb
saksham-jain: `context` would be better instead of `describe`. Isn't it?

joeldrapper (approved):

PR #36, opened 2022-07-19
Initial Rails helpers

joeldrapper:
[WIP] @enderahmetyurt and I paired on some initial helpers for Rails today, though I’m not entirely sure how we should best include them.

1. I don't want Phlex to be dependent on Rails, all the Rails integration should be optional.
2. I’m not sure if we should have a Railtie or Rails Engine, but we'll likely want to extend the application with some controller methods and custom helpers.

For now, we put both features in the same file.

### `render_page`
The idea is that this method would be available in a rails controller so if you used a consistent naming convention for your page components, you could render the correct page automatically.

```ruby
class ArticlesController < ApplicationController
  def index
    render_page articles: Article.all
  end
end
```

### `component`

The component method is designed to be used in ERB templates so you can render a Phlex component from your existing Rails views.

### DIFF
--- a/lib/phlex/rails.rb
+++ b/lib/phlex/rails.rb
@@ -1,0 +1,4 @@
+module Phlex
+  module Rails
+  end
+end
--- a/lib/phlex/rails/helpers.rb
+++ b/lib/phlex/rails/helpers.rb
@@ -1,0 +1,14 @@
+module Phlex
+  module Rails
+    module Helpers
+      def render_page(...)
+        page = "#{controller_name.classify}::#{action_name.classify}Page".constantize
+        render html: component(page, ...)
+      end
+
+      def component(component, ...)
+        component.new(...).call.html_safe
+      end
+    end
+  end
+end


alexandreruban:
@joeldrapper, maybe we could have two different gems: 

- `phlex` could contain the logic for creating components. This gem would be independent of any framework.
- `phlex-rails` would contain the Rails-specific modules and the Engine/Railtie.

It could be just like [factory_bot](https://github.com/thoughtbot/factory_bot) and [factory_bot_rails](https://github.com/thoughtbot/factory_bot_rails/tree/main/lib/factory_bot_rails). What do you think?

joeldrapper:
@alexandreruban that's a really good option. The alternative is to have a `rails.rb` file that can be optionally required. I think you could just do this in the Gemfile.

```ruby
gem "phlex", require: "phlex/rails"
```

joeldrapper:
I think it would be easier to extract a `phlex-rails` gem later than it would be to merge a separate gem back into the main one, so my instinct is to see how far we can get with just one gem while always keeping that door open. I don't feel strongly about this though.

PR #37, opened 2022-07-20
Support `<template>` elements

joeldrapper:
Previously, you could not use `<template>` elements in your templates because of a naming conflict. Calling the `template` method result in an infinite loop.

We overcome this here by prepending a `template` method that checks if it’s already rendering or not. If a render is already in progress, `template` acts as the DSL method, otherwise calls `super`.

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -18,6 +18,16 @@ module Phlex
         return super unless @_cache
         Rails.cache.fetch(self) { super }
       end
+
+      def template(...)
+        if @_rendering
+          _template_tag(...)
+        else
+          @_rendering = true
+          super
+          @_rendering = false
+        end
+      end
     end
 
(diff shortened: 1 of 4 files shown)


alexandreruban (approved):
Nice trick! That works well for me 🥳

I would have tried a different approach, as I would prefer not to treat the `<template>` tag differently from others. What I would have tried is to evaluate the content of the `Component#template` method in another context:

```rb
class Phlex::Component
  # ... 
  
  def call
    SomeOtherClassThatWeCouldCallContext.instance_eval { template }
  end
end
```

Does it make sense?

I'm not sure it's worth exploring though as your implementation already works well 😃

joeldrapper:
I don't think that approach would have worked as it's not possible to `instance_exec` a bound method against another object like you can other callables.

PR #38, opened 2022-07-21
Support custom elements

### LINKED ISSUE #33: Support custom elements
I'm excited to see how this library evolves! One thing I was wondering is the ability to support custom element tags (aka arbitrary tag names with at least one dash).

I worked on a [PR for Papercraft](https://github.com/digital-fabric/papercraft/pull/2) to support converting Ruby calls like `my_custom_element` to `my-custom-element` and would be happy to contribute here if you have an approach in mind.

joeldrapper:
An initial prototype for supporting the definition of custom elements. Here's an example of how you might use it to build a Trix component:

```ruby
class TrixComponent < Phlex::Component
  register_element :trix_editor

  def template
    input id: "content", type: "hidden"
    trix_editor input: "content", role: "textbox"
  end
end
```

Custom elements, for now, are local to the component where they're registered.

Closes #33

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -34,6 +34,20 @@ module Phlex
     include Node, Context
 
     class << self
+      def register_element(*tag_names)
+        tag_names.each do |tag_name|
+          unless tag_name.is_a? Symbol
+            raise ArgumentError, "Custom elements must be provided as Symbols"
+          end
+
+          class_eval(<<-RUBY, __FILE__, __LINE__ + 1)
+            def #{tag_name}(...)
+              _standard_element("#{tag_name.to_s.gsub('_', '-')}", ...)
+            end
+          RUBY
+        end
+      end
+
       def inherited(child)
         child.prepend(Overrides)
         super
--- a/spec/phlex/component_spec.rb
+++ b/spec/phlex/component_spec.rb
@@ -298,5 +298,37 @@ RSpec.describe Phlex::Component do
         expect(output).to eq "<div><h1>A</h1><template>B</template><h2>C</h2></div>"
       end
     end
+
+    describe "with custom elements" do
+      let :component do
+        Class.new Phlex::Component do
+          register_element :trix_editor, :trix_toolbar
+
+          def template
+            div {
+              trix_toolbar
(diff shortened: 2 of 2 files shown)


lib/phlex/component.rb
```diff
@@ -34,6 +34,18 @@ def template(...)
     include Node, Context
 
     class << self
+      def register_element(tag_name)
+        unless tag_name.is_a? Symbol
+          raise ArgumentError, "Custom elements must be provided as Symbols"
+        end
```
joeldrapper, line 40: It's a requirement for custom elements to be comprised of two or more words, so we could add a check here.

alexandreruban:
That looks great!

Just a small question, wouldn't it be possible to add this behavior directly on the `Phlex` module and add the custom element to the `Phlex::Tag::StandardElement::ELEMENTS` array?

Something like this: 

```rb 
module Phlex
  def self.register_element(tag_name)
    Phlex::Tag::StandardElement::ELEMENTS << tag_name
  end
end
```

The limitation is that once we load the `Phlex::Component` class, we can't add more custom elements as the `class_eval` already run. I'm not sure it would work though depending on the order the classes are read.

joeldrapper:
@alexandreruban we could definitely support global custom elements like that, but I wanted to first look at supporting custom elements specifically scoped to only the component that registered them. I think in practice, this will be sufficient. But still open to globals if we can think of a use-case.

lib/phlex/component.rb
joeldrapper: I think they must also be lowercase but we could do a transform to ensure that.

lib/phlex/component.rb
```diff
@@ -34,6 +34,18 @@ def template(...)
     include Node, Context
 
     class << self
+      def register_element(tag_name)
```
joeldrapper, line 37: We should take a splat of tag names for defining multiple custom elements.

joeldrapper:
It's worth noting that you can register custom elements in an abstract component class, e.g. `ApplicationComponent` and then subclass that with your other components.

alexandreruban (approved):
LGTM! 🎉🎉

PR #39, opened 2022-07-21
Raw node type

joeldrapper:
I've added a raw node type that doesn't escape HTML.

Usage:

```ruby
class ExampleComponent < Phlex::Component
  def template
    _raw "<h1>Hello</h1>"
  end
end
```

The underscore suffix is to mark this is a private method that you shouldn't really use.

I also added a spec for text nodes.


alexandreruban (approved):
Super clear! 👍

PR #40, opened 2022-07-21
Whitespace node

joeldrapper:
Added a `whitespace` node for when you need to insert whitespace between two elements. For example:

```ruby
class ExampleComponent < Phlex::Component
  def template
    a "Home"
    whitespace
    a "About"
  end
end
```


alexandreruban (approved):
And super clear again! 😄

joeldrapper:
Thanks for reviewing so quickly! 🙌

Issue #41, opened 2022-07-22
Components should be streamable in Rails


joeldrapper:
It looks like streaming in Rails is handled here. 
https://github.com/rails/rails/blob/04972d9b9ef60796dc8f0917817b5392d61fcf09/actionpack/lib/action_controller/metal/streaming.rb#L195-L220

It takes a [`Rack::Chunked::Body`](https://rdoc.info/github/rack/rack/main/Rack/Chunked/Body) which takes an enumerator. Rack made the [decision](https://github.com/rack/rack/issues/1266#issuecomment-523226433) to depreciate this module and delegate it to the server.

joeldrapper:
I think this is mostly made possible with #42, but I’m the issue open until we've conducted further testing. I don't know if we need some kind of batch chunker to get the best performance from HTTP streaming. Never even looked into it before.

joeldrapper:
I’m closing this as I don't have any plans to revisit streaming at this time. At least not until Rails / Rack / Puma figure out who's responsible for what. Phlex works off a simple string buffer, and its only requirement is that the buffer respond to `<<`.

PR #42, opened 2022-07-24
Use fewer strings

joeldrapper:
Elements pass a buffer string around and append to it instead of returning their own strings for joining.

In theory, this means we can yield to an enumerator for streaming responses. (See #41)

```ruby
Enumerator.new { |buffer| component.call(buffer) }
```

### DIFF
--- a/lib/phlex/block.rb
+++ b/lib/phlex/block.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Phlex
   class Block
     include Callable
--- a/lib/phlex/callable.rb
+++ b/lib/phlex/callable.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Phlex
   module Callable
     def call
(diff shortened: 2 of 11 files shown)


joeldrapper:
This appears to be slightly slower compared with `Array#join`, but it should be faster once Jean's [improvements](https://twitter.com/_byroot/status/1550119824524599298) to `String#<<` are merged.

joeldrapper:
Correction, this change actually does improve performance, but should be even better in newer Rubies where `String#<<` is optimised.

Before
```
ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [arm64-darwin21]
        NavComponent     3.903k i/100ms
        NavComponent     39.907k (± 1.2%) i/s -    202.956k in   5.086515s
```
After
```
ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [arm64-darwin21]
        NavComponent     4.175k i/100ms
        NavComponent     41.229k (± 1.4%) i/s -    208.750k in   5.064122s
```

alexandreruban (approved):
Nice!! 😄

PR #45, opened 2022-07-25
Remove fragment caching

### LINKED ISSUE #2: Fragment caching
The cache key should include

- [x] User provided resource
- [x] Ruby version
- [x] Phlex version
- [x] Content block definition checksum
- [x] Relevant ancestors checksums
- [x] Sub component checksums
- [ ] Sub component checksums from the content block definition

joeldrapper:
We’re not going to need fragment caching with compiled templates. Any caching we do still need will be only caching resources, such as ActiveRecord relations, so almost all of this stuff will be irrelevant.

Let's remove it for now and we can always refer back to this PR if we need to add anything back.

Closes #2


lib/phlex/component.rb
```diff
@@ -121,34 +71,6 @@ def render_block(new_target, ...)
...
-
-    def cacheable_resources
-      (@_cache == true) ? assigns : @_cache
-    end
-
-    def cacheable_content
-      CacheableObject.new(@_content)
-    end
-
     def assigns
```
alexandreruban, line 74: I think you can remove this method as well :)

joeldrapper: I left this in because it's quite nice for meta programming components. That said, I'd have a hard time coming up with a real use-case so maybe it's best to remove it for now. It's easy enough to put back in.

davidalejandroaguilar:
@joeldrapper Hey there, I'm just following these issues around and wondering if Phlex works with Rails fragment caching. I've searched issues on this and the phlex-rails repo, as well as the READMEs and website and can't seem to get an answer.

PR #46, opened 2022-07-25
Strip `javascript:` from href attribute of anchor elements

### LINKED ISSUE #5: Block javascript protocol in links
If a rails app renders a link from user input, it might be possible for a user to run arbitrary JavaScript on some else's account by using the `javascript:` link protocol. To guard against this Cross Site Scripting (XSS) attack, we should strip `javascript:` from the beginning of the `href` attribute on `<a>` tags.

saksham-jain:
- Strip `javascript:` from `href` attribute of `anchor` elements

Closes #5

### DIFF
--- a/lib/phlex/tag.rb
+++ b/lib/phlex/tag.rb
@@ -45,7 +45,7 @@ module Phlex
       attributes = @attributes.dup
       attributes[:class] = classes
       attributes.compact!
-      attributes[:href] = attributes[:href].gsub("javascript:", "") if attributes[:href]
+      attributes[:href].sub!(/^\s*(javascript:)+/, "") if attributes[:href]
       attributes.transform_values! { ERB::Util.html_escape(_1) }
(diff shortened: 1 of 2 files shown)


lib/phlex/tag.rb
```diff
@@ -45,6 +45,7 @@ def attributes
       attributes = @attributes.dup
       attributes[:class] = classes
       attributes.compact!
+      attributes[:href] = attributes[:href].gsub("javascript:", "") if attributes[:href]
```
joeldrapper, line 48: Let's use a regular expression to only substitute this if it's the start of the string `^\s*(javascript:)+`. Given this expression, we could use `sub` instead of `gsub`. We can also mutate the original string rather than create a new string with `sub!`.

```suggestion
      attributes[:href].sub!(/^\s*(javascript:)+/, "") if attributes[:href]
```

joeldrapper (commented):
Thanks for the PR. I added one suggestion but otherwise looks great!

PR #48, opened 2022-07-25
Strip `javascript:` protocol only after HTML escaping

joeldrapper:
I realised it _might_ be possible to format a string such that that it is not stripped by our regular expression but the HTML escaper upgrades it. If the html escaper converts some character to a `:` to prevent some other attack, the order of these operations could be exploited.

For example `javascript：alert(1)` will not stripped by our regular expression because `：` is not `:`. If the HTML escaper converts `：` to `:` then it will become `javascript:alert(1)`.

I haven't been able to find an actual workaround to our existing setup so this is just precautionary.

### DIFF
--- a/lib/phlex/tag.rb
+++ b/lib/phlex/tag.rb
@@ -41,8 +41,8 @@ module Phlex
       attributes = @attributes.dup
       attributes[:class] = classes
       attributes.compact!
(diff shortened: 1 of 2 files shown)


alexandreruban (approved):
Makes sense! LGTM! 🎉

PR #50, opened 2022-07-25
Optimize delegation

byroot:
For _void_element we can directly use `alias_method` and then use `__callee__.name` to know which tag to generate.

This way it saves a method call, with `...` delegation which isn't that fast.

Unfortunately `_standard_element` can't be delegated this way because of the `template` method. However we can enable `frozen_string_literal: true` in `class_eval` to save a string allocation per call.

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -48,14 +48,15 @@ module Phlex
       Tag::ClassCollector.new(self, tag)
     end
 
-    def _void_element(name, **kwargs)
-      tag = Tag::VoidElement.new(name, **kwargs)
+    def _void_element(**kwargs)
+      tag = Tag::VoidElement.new(__callee__.name, **kwargs)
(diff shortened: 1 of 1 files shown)


joeldrapper:
Thanks @byroot! 🎉  I wonder if we should switch back to inlining the the standard_element method definitions, like it was before #37.

byroot:
I dunno, it depends on how big `Tag::StandardElement::ELEMENTS` is. If it's really large it means a lot of identical code, so more cache misses etc. 

That's why the `alias_method + __calle__` combo usually gives the best result both in perf and in code size. But it's a bit more constrained.

It also depends on whether that `template` conflict could be resolved differently, but I don't know your gem much, I was just skimming the code and saw this.

joeldrapper:
@byroot I was able to get your technique to work with `_standard_element` in #52.

PR #51, opened 2022-07-25
Rename `Node` -> `ParentNode`

joeldrapper:
Node doesn't feel like the right name for this as not all nodes have children so they don't all include `Node`. I’m not entirely sure if `Branch` is the right name either so let me know if you can think of anything else.

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -21,7 +21,7 @@ module Phlex
       end
     end
 
-    include Node, Context
+    include ParentNode, Context
 
     class << self
       def register_element(*tag_names)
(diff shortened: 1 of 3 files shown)


fractaledmind:
`ParentNode`?

alexandreruban:
I liked `Node`, it made me think of a `DOM node`. I understand the fact that they always have children though.

Maybe `ParentElement` like in the `DOM` would make sense? A `ParentElement` has children? I can't think of a better option right now 😅

joeldrapper:
“Parent” is good. I think `ParentNode` is slightly better than `ParentElement` because these parent nodes aren't necessarily HTML elements, they could be components.

PR #52, opened 2022-07-25
Optimize delegation

joeldrapper:
@byroot opened a PR (#50) to optimise delegating by using `alias_method` with `__callee__.name` instead of delegating with `...`. Building on that, I found a way to apply it to the `_standard_elements` by making the `name` argument optional. This optimisation makes the nav benchmark ~3% faster.

Before

```
        NavComponent     4.441k i/100ms
        NavComponent     44.677k (± 0.4%) i/s -    226.491k in   5.069581s
```

After

```
        NavComponent     4.593k i/100ms
        NavComponent     46.011k (± 0.4%) i/s -    234.243k in   5.091106s
```

Closes #50

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -48,14 +48,15 @@ module Phlex
       Tag::ClassCollector.new(self, tag)
     end
 
-    def _void_element(name, **kwargs)
-      tag = Tag::VoidElement.new(name, **kwargs)
+    def _void_element(**kwargs)
+      tag = Tag::VoidElement.new(__callee__.name, **kwargs)
       self << tag
       Tag::ClassCollector.new(self, tag)
     end
 
     Tag::StandardElement::ELEMENTS.each do |tag_name|
       class_eval(<<-RUBY, __FILE__, __LINE__ + 1)
+        # frozen_string_literal: true
         def #{tag_name}(...)
           _standard_element("#{tag_name}", ...)
         end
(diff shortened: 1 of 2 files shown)


lib/phlex/context.rb
```diff
@@ -26,12 +26,14 @@ def component(component, *args, **kwargs, &block)
...
-      _standard_element("template", ...)
+    def _template_tag(*args, **kwargs, &)
+      _standard_element(*args, _name: "template", **kwargs, &)
     end
 
-    def _standard_element(name, content = nil, **kwargs, &block)
+    def _standard_element(content = nil, _name: nil, **kwargs, &block)
       raise ArgumentError if content && block_given?
+
+      name = _name ||= __callee__.name
```
alexandreruban, line 36: Do we need this `name` variable here? Maybe `_name ||= __callee__.name` would be sufficient what do you think?

alexandreruban (approved):
Nice trick! TIL about `__callee__`!

I don't know if it's intended, but you have Jean's commit on this branch.

lib/phlex/context.rb
joeldrapper: No need for it, I just think we may as well since we then don't need to use the underscored variable name later. AFAIK local variable assignment is optimised away anyway so this shouldn't have a performance impact.

joeldrapper:
> I don't know if it's intended, but you have Jean's commit on this branch.

Yep, this is building on his work in #50.

joeldrapper:
Thanks @byroot this is a really neat trick.

lib/phlex/context.rb
byroot: > AFAIK local variable assignment is optimised away anyway 

No it's not. Local variables can be inspected through `binding` etc, so Ruby can't eliminate them.

They don't have a big cost though, one more local variable won't make a noticeable difference.

PR #53, opened 2022-07-25
Use UTF-8 buffers

byroot:
`String.new` isn't a good way to create a buffer, or at least not for HTML templating which is likely to produce UTF-8.

`String.new` creates a binary string:

```ruby
>> String.new.encoding
=> #<Encoding:ASCII-8BIT>
```

Since your final string should be UTF-8, it's better to do the buffering with UTF-8 strings.


joeldrapper:
TIL! 😍

PR #59, opened 2022-07-26
Treat component's first positional argument as text

### LINKED ISSUE #56: Component text content
The first positional argument given to a component should be treated as its text content if no block is given and the component doesn't itself handle the positional argument.

Given this component

```ruby
class HeaderComponent < Phlex::Component
  def template(&)
    h1.text_xl(&)
  end
end
```

We should be able to use it like this.

```ruby
component HeaderComponent, "Hello World!"
```

alexandreruban:
Closes #56

Thanks to this commit, given this component:

```rb
class HeaderComponent < Phlex::Component
  def template(&)
    h1.text_xl(&)
  end
end
```

We are able to use it like this:

```rb
component HeaderComponent, "Hello World!"
```

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -23,7 +23,11 @@ module Phlex
         block = Block.new(self, &block)
       end
 
-      self << component.new(*args, **kwargs, &block)
+      if args.one? && !block_given?
+        self << component.new(**kwargs) { text(args.first) }
+      else
+        self << component.new(*args, **kwargs, &block)
+      end
     end
 
     def _template_tag(*args, **kwargs, &)
--- a/spec/phlex/component_spec.rb
+++ b/spec/phlex/component_spec.rb
@@ -405,4 +405,18 @@ RSpec.describe Phlex::Component do
       end
     end
   end
+
+  describe "with a positional argument" do
+    let(:component) do
+      Class.new Phlex::Component do
+        def template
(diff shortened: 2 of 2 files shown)


joeldrapper:
I wonder if there's a way to use `Method#parameters` to check if I've defined a component that takes a positional argument.

```ruby
class LinkComponent
  def initialize(text, to:)
    @text = text
    @url = to
  end

  def template
    a.underline.text_sky_500 @text, href: @url
  end
end
```

I know we could re-write this component like to use the content block and it would be even more flexible, but it might be nice to not require authors of components know this.

```ruby
class LinkComponent
  def initialize(to:)
    @url = to
  end

  def template(&)
    a.underline.text_sky_500 href: @url, &
  end
end
```

This second component also doesn't require its content, we should work out how to do that.

alexandreruban:
Oh OK, I understand better now, I'll update my PR soon!

joeldrapper:
I don't know if it'll be possible — we override that method so it might be difficult to get to the original — and if not, I think what you've done here is really good.

joeldrapper (approved):
@alexandreruban shall we just ship this as it is and then layer on parameter checking later if we can?

alexandreruban:
@joeldrapper yes we can! I had a look at `Method#parameters` but couldn't find a way to make it work.

PR #61, opened 2022-07-28
Fix CSS class with frozen string

### LINKED ISSUE #60: Magic frozen string comment breaks component
Hello, 

I started migrating some ViewComponent files to Phlex but I am getting "can't modify frozen String" whenever I try to include a class attribute to an anchor tag. 

```ruby
# app/frontend/navbar/component.rb

# frozen_string_literal: true

module Navbar
  class Component < Phlex::Component
    def template
      nav.navbar.bg_base_100 do
        a("Get started", href: Current.rodauth.login_path, class: "btn")
      end
    end
  end
end
```

After checking out this [file](https://github.com/joeldrapper/phlex/blob/main/lib/phlex/tag.rb#L23), I realized that I needed to remove the magic comment. I think it should be added in the README or are they any plans to support it?

Also, I think it would be nice to have a section to show how to embed a Phlex component in an erb template for starters. It's getting very late over here and  It took me more than I am willing to admit to come up with: `<%= render html: Navbar::Component.new.call.html_safe %>` :sleepy: :joy:

joeldrapper:
Closes #60 and also improves performance.

Before:
```
        NavComponent     4.648k i/100ms
        NavComponent     46.367k (± 0.4%) i/s -    232.400k in   5.012298s
```

After:
```
        NavComponent     4.644k i/100ms
        NavComponent     46.789k (± 0.2%) i/s -    236.844k in   5.062034s
```

### DIFF
--- a/lib/phlex/tag.rb
+++ b/lib/phlex/tag.rb
@@ -11,7 +11,7 @@ module Phlex
 
     def initialize(name, **attributes)
       @name = name
-      @classes = String.new
+      @classes = +""
       self.attributes = attributes
     end
 
@@ -23,11 +23,11 @@ module Phlex
     def classes=(value)
       case value
       when String
-        @classes << value.prepend(SPACE)
+        @classes << SPACE << value
       when Symbol
-        @classes << value.to_s.prepend(SPACE)
+        @classes << SPACE << value.to_s
       when Array
-        @classes << value.join(SPACE).prepend(SPACE)
+        value.each { @classes << SPACE << _1 }
       when nil
         return
       else
--- a/spec/phlex/component_spec.rb
+++ b/spec/phlex/component_spec.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class CardComponent < Phlex::Component
   def template(&)
     article.p_5.rounded.drop_shadow(&)


spec/phlex/component_spec.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
joeldrapper, line 1: Adding this caused our specs to fail for the same reason described in #60.

lib/phlex/tag.rb
```diff
@@ -11,7 +11,7 @@ class Tag
 
     def initialize(name, **attributes)
       @name = name
-      @classes = String.new
+      @classes = +""
```
joeldrapper, line 14: We know this will be a UTF-8 String. See #53

lib/phlex/tag.rb
```diff
@@ -23,11 +23,12 @@ def attributes=(attributes)
...
       when String
-        @classes << value.prepend(SPACE)
+        @classes << SPACE << value
       when Symbol
-        @classes << value.to_s.prepend(SPACE)
+        @classes << SPACE << value.to_s
       when Array
-        @classes << value.join(SPACE).prepend(SPACE)
+        @classes << SPACE
+        value.each { @classes << _1 }
```
joeldrapper, line 31: `String#<<` is faster than `Array#join` and will be even faster in future Rubies.

lib/phlex/tag.rb
```diff
@@ -23,11 +23,11 @@ def attributes=(attributes)
...
       case value
       when String
-        @classes << value.prepend(SPACE)
+        @classes << SPACE << value
       when Symbol
-        @classes << value.to_s.prepend(SPACE)
+        @classes << SPACE << value.to_s
       when Array
-        @classes << value.join(SPACE).prepend(SPACE)
+        value.each { @classes << SPACE << _1 }
```
joeldrapper, line 30: `String#<<` is faster than `Array#join` and will be even faster in future Rubies.

alexandreruban (approved):

PR #64, opened 2022-07-28
Cache CSS classes on ClassCollector

### LINKED ISSUE #54: Cache CSS classes in the ClassCollector
The `ClassCollector` uses `method_missing` to pick up arbitrary CSS classes chained on element methods. Chances are, you're going to use the same CSS classes more than once, so I wonder if it might make sense to cache a method on the ClassCollector to avoid the `missing_method` call the next time that CSS class is used.

I tried to put together a PoC but it's tricky because you can't use things like `send` or `self.class.define_method` from a `BasicObject`.

One possible drawback is that this could lead to a memory leak if someone sends arbitrary classes (think UUID or something weird) using `send` from a component. We could keep a counter of cached CSS classes and make sure it doesn't grow too big.

joeldrapper:
This PR improves performance of the `ClassCollector` when a CSS class is re-used. Closes #54 

### How it works

The `ClassCollector` picks up CSS classes through method chaining on an element. For example, if we want to add the CSS classes `text-xl` and `font-bold` to an `<h1>` tag, we could write the following:

```ruby
h1.text_xl.font_bold "Hello World!"
```

Because there's no way to know ahead-of-time that `text_xl` and `font_bold` will be called on the `h1`, these method calls are picked up through `method_missing`. Unfortunately, `method_missing` is quite slow so now we make a new alias for that missing method the first time it's called so it's not missing the next time if the same CSS class is used elsewhere.

We don't have a benchmark that really showcases this, but it definitely improves performance when CSS classes are used more than once with method chaining.

### DIFF
--- a/lib/phlex/tag/class_collector.rb
+++ b/lib/phlex/tag/class_collector.rb
@@ -8,21 +8,29 @@ module Phlex
         @tag = tag
       end
 
-      def method_missing(name, content = nil, **attributes, &block)
-        @tag.classes = name
+      define_method :__class__, ::Kernel.instance_method(:class)
+      define_method :__callee__, ::Kernel.instance_method(:__callee__)
+
+      def __add_class__(content = nil, **attributes, &block)
+        @tag.classes = __callee__.name
         @tag.attributes = attributes if attributes
 
         if ::Kernel.block_given?
           if block.binding.receiver.is_a?(Block)
             block.call(@tag)
           else
-            Block.new(@context, &block).call(@tag)
+            ::Phlex::Block.new(@context, &block).call(@tag)
           end
         end
 
-        Block.new(@context) { text content }.call(@tag) if content
+        ::Phlex::Block.new(@context) { text content }.call(@tag) if content
 
         self
+      end
+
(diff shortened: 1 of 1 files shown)


yogeshjain999:
Nice trick with `__callee__`

Wondering if something similar is planned for element's ID or other attributes.

joeldrapper:
> Wondering if something similar is planned for element's ID or other attributes.

I don't think we can optimise those since they're always provided as arguments not method calls.

alexandreruban (approved):
Nice trick!! 🤯

yogeshjain999:
> > Wondering if something similar is planned for element's ID or other attributes.
> 
> I don't think we can optimise those since they're always provided as arguments not method calls.

hmm understood! One more doubt 🙃 - how can we add classes conditionally in this behaviour ?

joeldrapper:
You can't. Best way to add classes conditionally is to pass them in as the `class` keyword argument. If the conditional logic is complex, you can extract a method.

yogeshjain999:
Yeah agree. I guess then chaining classes would not be ideal in all the cases, in case if we decide to define a standard format of writing them.

PR #68, opened 2022-07-29
Enable partial rendering from components

alexandreruban:
References #1 

### Phlex/Rails integration

This PR allows Rails users to render partials from Phlex components.

### Usage

In the `Gemfile` of the Rails application, we need to require the additional `phlex/rails` file to enable Phlex's Rails extensions: 

```rb
# Gemfile

gem "phlex", require: ["phlex", "phlex/rails"]
```

Now, let's imagine this `PagesController` that renders the HTML produced by the the `HomePage` component:

```rb
class PagesController < ApplicationController
  def home
    @name = "Alexandre"

    render html: HomePage.new(name: @name).call.html_safe
  end
end
```

The `HomePage` component can now render partials: 

```rb
class HomePage < Phlex::Component
  def template(&)
    h1 "Home page"

    render partial: "pages/content", locals: { name: @name }
  end
end
```

See the [dummy application](https://github.com/alexandreruban/phlex-dummy-app/commit/cc909f7110407499254048b5575970df8200d901) for reference.

### Todo

- [ ] Add specs
- [ ] Add documentation already?
- [x] Use (...) instead of (*args, **kwargs, &block)


alexandreruban:
@joeldrapper any idea how to test this properly without a `test/dummy` application?

lib/phlex/rails.rb
```diff
@@ -0,0 +1,6 @@
+require "phlex/rails/rendering"
```
joeldrapper, line 1: If require `phlex` here, everything else should be auto-loaded and we won't need to use an Array require in the Gemfile.

lib/phlex/rails/rendering.rb
```diff
@@ -0,0 +1,11 @@
+module Phlex
+  module Rails
+    module Rendering
+      def render(...)
+        _raw ApplicationController.render(...)
```
joeldrapper, line 5: I expect we can use [`ActionController::Rendering#render_to_string`](https://api.rubyonrails.org/classes/AbstractController/Rendering.html#method-i-render_to_string) for this.

lib/phlex/rails/rendering.rb
```diff
@@ -0,0 +1,11 @@
...
+      end
+    end
+  end
+end
+
+Phlex::Context.send(:include, Phlex::Rails::Rendering)
```
joeldrapper, line 11: Do we need to `send` this or can we just call `include` directly?

```suggestion
Phlex::Context.include(Phlex::Rails::Rendering)
```

joeldrapper:
> any idea how to test this properly without a test/dummy application?

We should be able to add Rails as a development dependency and then create a partial in `examples`. Not 100% sure how the partial lookup works in Rails, but I expect we can get it to look at our partial, even if we need to use a Mocha stub.

fractaledmind:
One middle ground approach I have used for testing Rails integration for a gem is the [cumbustion gem](https://github.com/pat/combustion). From the start of the REAMDE:
>Combustion is a library to help you test your Rails Engines in a simple and effective manner, instead of creating a full Rails application in your spec or test folder.
>
>It allows you to write your specs within the context of your engine, using only the parts of a Rails app you need.

spec/phlex/component_spec.rb
```diff
@@ -421,4 +421,22 @@ def template
...
+        def template
+          component CardComponent do
+            render "content", locals: { name: "Alexandre" }
+          end
+        end
+      end
+    end
+
+    it "produces the correct markup" do
+      ActionView::Base.with_view_paths(File.expand_path("../../examples/partials", __dir__)) do
```
alexandreruban, line 437: Maybe we should do this globally and not per spec.
I'll have a look if I can find a option that does that.

joeldrapper: You'll be able to do it with an around each I think.

joeldrapper:
Hey @alexandreruban I've just taken this PR for a spin. It works an it's awesome! Couple of things:

1. I wonder if we can find a way to either use a root namespace class like our own `ApplicationControllerForPhlex` or something like that in order to support looking up partials by model, e.g. `render @article`.
2. Within an ActionView view, you can render a partial with the partial as the first argument rather than a `partial:` keyword argument. I wonder if we can support that too?

spec/phlex/component_spec.rb
alexandreruban: I couldn't find a way to make this work. The following code does not work for instance:

```rb
around(:each) do |example|
  ActionView::Base.with_view_paths(...) do 
    example.run
  end
end
```

alexandreruban:
Hey @joeldrapper I'm not sure to understand your last comment:

> I wonder if we can find a way to either use a root namespace class like our own ApplicationControllerForPhlex or something like that in order to support looking up partials by model, e.g. render `@article`.

I added one spec with `render @article` inside a Phlex component and it works as expected: 

```rb
Class.new Phlex::Component do
  def template
    component CardComponent do
      render @article
    end
  end
end
```

> Within an ActionView view, you can render a partial with the partial as the first argument rather than a partial: keyword argument. I wonder if we can support that too?

There is also an example of this behavior in the specs: 

```rb
Class.new Phlex::Component do
  def template
    component CardComponent do
      render "content", locals: { name: "Alexandre" }
    end
  end
end
```

I'm probably missing something here, feel free to add a commit directly to this PR if you have some ideas I miss :)

joeldrapper:
In my manual tests it searches for partials in the `action_controller` directory because `ActionController::Base` is namespaced. It seems to use the controller namespace to determine the search path for partials.

joeldrapper:
Closing in favour of #70

PR #69, opened 2022-08-02
Implement Phlex::Component#render_in

grncdr:
References #1, #68

Allows Phlex::Component instances to be passed to the `render` view helper in Rails.

Currently this is just an alias for `#call` with no arguments, and as such it completely ignores the view context argument.

However, there was some discussion on Discord about maybe supporting capturing blocks like:

```rb
<%= render MyComponent.new do %>
  other ERB content
<% end %>
```

This is not super hard to implement if there's interest, but I'm not sure if it will conflict or be confused with other, more idiomatic, uses of passing blocks to Phlex components. (e.g. the table fabricator example).

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -55,6 +55,10 @@ module Phlex
       template(&@_content)
       super
     end
+    
+    def render_in(_view_context)
+      call
+    end
 
     def target
       @_target || self


grncdr:
The ActionController API has an (apparently undocumented) larger contract. The renderable "thing" passed to `#render` should _also_ respond to `:format`. I believe this is used for setting a content type header.

I did a brief test "shot in the dark" test by adding:

```rb
def format
  :html
end
```

And that seemed to make `render MyComponent.new` work from controllers, but I haven't figured out where in Rails that's called and if `:html` is actually the right thing to return.

joeldrapper:
I had no idea it would be this simple. Thanks for following up on this. 🙏 

> This is not super hard to implement if there's interest, but I'm not sure if it will conflict or be confused with other, more idiomatic, uses of passing blocks to Phlex components. (e.g. the table fabricator example).

I would love it if we could support both. There must be something unique about the block that's captured from ERB after the `%>`.

grncdr:
> I would love it if we could support both. There must be something unique about the block that's captured from ERB after the `%>`.

There really isn't much special about it, it's just a normal block passed to `#render_in` that will eval in the ERB context (e.g. strings are appended to the output buffer). You can capture the output of the block using `view_context#capture` so `#render_in` can look like this:

```rb
def render_in(view_context, &block)
  raise "too many blocks 😅" if block && @_content
  @_content = proc { Raw.new(view_context.capture(&block)) } if block
  call
end
```

I don't know if that quite works, since I haven't actually read enough of Phlex's code to say if `Raw` is going to do what I expect. The important thing is that `view_context.capture(&block)` returns a string that you can then nest somewhere inside the phlex output.

However, if my assumptions about `Raw` are correct, that would let you do something like:

```rb
# copied from the README
class CardComponent < Phlex::Component
  def initialize(title:)
    super
  end

  def template(&)
    div.rounded.drop_shadow.p_5 do
      h1 @title
      content(&)
    end
  end
end
```

```erb
<%= render Card.new(title: "Nested ERB content") do %>
  <figure>
      <blockquote cite="https://github.com/joeldrapper/phlex/pull/69">
          <p>Weird phlex, but ok.</p>
      </blockquote>
      <figcaption>—<%= current_user.name %></figcaption>
  </figure>
  <%= current_user.name %>
<% end %>

**Comment by grncdr** [comment]
I submitted my previous comment too soon. I meant to clarify that this is where the conflict arises. A normal "phlex block" is normally passed to `#initialze`, so to use it with `render` you need to add parens:

```erb
<%= render(CardComponent.new(title: "Nested Phlex content") do
  p "not the prettiest..."
end) %>
```

If you want to try and distinguish between "block types" (is this a phlex or ERB block?) passed to `#render_in` I _suspect_ the performance will fall off a very cliff, but I've been wrong about this sort of thing before. :wink:

joeldrapper:
I’ve almost got this working in a branch, but ERB is outputting the ERB content inside the block.

grncdr:
Interesting, I assumed you need to evaluate the block somewhere for that to happen. I was testing out the behaviour with this very simple "component":

```rb
  class HelloWorldComponent
    def render_in(view_context, &block)
      <<~HTML.html_safe
        <div class="hello-world">
          Hello world. Your content is:<br>
          #{view_context.capture(&block) if block}
        </div>
      HTML
    end
  end
```

```erb
<%= render HelloWorldComponent.new do %>
<strong>really cool</strong>
<% end %>
```

output:

```html
<div class="hello-world">
  Hello world. Your content is:<br>
  
    <strong>really cool</strong>

</div>
```

joeldrapper:
I’m getting really close. Now have the second "Hello World" in the CardComponent.

joeldrapper:
I can't figure out where the first one comes from.

joeldrapper:
I’ve done it!! Thank you so much for your help @grncdr. I’m going to open a new PR with your stuff @alexandreruban stuff and some extra stuff.

joeldrapper:
Closing in favour of #70

PR #70, opened 2022-08-02
Support Rails rendering

joeldrapper:
WIP combining the work from #69 and #68.

I’ve also added the ability to capture ERB yielded into a Phlex component.

### Initial Rails Support
Just update your `Gemfile` to require `phlex/rails`.

```ruby
gem "phlex", require: "phlex/rails"
```

### Rendering a Rails partial in a component
You can now render a Rails partial inside a component.

```ruby
class Article::IndexPage < Phlex::Component
  def initialize(articles:)
    super
  end

  def template
    ul do
      @articles.each do |article|
        li do
          render article # renders the partial views/articles/_article.html.erb
        end
      end
    end
  end
end
```

### Rendering a Phlex Component from a Rails ERB view
You can also render a Phlex component from an ERB view and even pass ERB content into it.

```erb
<%= render CardComponent.new do %>
  <h1>Hello world!</h1>
<% end %>
```

Here the block is passed to the `render` helper method rather than the `new` method on `CardComponent`. You can write your component content in Phlex instead by passing the block to the `new` method. The easiest way to do this is to use a brace block.

```erb
<%= render CardComponent.new {
  h1 "Hello world!"
} %>
```

### Rendering a Phlex Component from a Rails Controller

```ruby
class ArticlesController
  def index
    render Article::IndexPage.new(articles: Article.all)
  end
end
```


grncdr (approved):
Very nice, thanks taking the time to make all my half-baked code samples actually work. Really exciting stuff.

lib/phlex/rails/context.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+ApplicationControllerForPhlex = Class.new(ActionController::Base)
```
joeldrapper, line 3: It’s necessary to have a controller at the root namespace otherwise ActionView searches for partials inside the folder that matches that namespace of the controller.

lib/phlex/rails/context.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Phlex
+  module Rails
+    module Context
+      def render(partial = nil, **kwargs)
```
joeldrapper, line 8: I think capturing the first argument as the `partial` and soaking up `kwargs` as the locals is more consistent with the `render` view helper in Rails.

lib/phlex/component.rb
```diff
@@ -6,7 +6,12 @@ module Phlex
   class Component
     module Overrides
       def initialize(*args, **kwargs, &block)
+        if block_given? && !block.binding.receiver.is_a?(Block)
+          block = Block.new(self, &block)
+        end
```
joeldrapper, line 11: We need to ensure these blocks are Phlex::Blocks here. This doesn't mean we can remove the Phlex::Block coercion from the other context methods, since it's important to coerce immediately to capture the right context for the Phlex::Block.

jaredcwhite:
@joeldrapper I've gotten a variation of this working in Bridgetown as well!

Repo: https://github.com/jaredcwhite/test-bt-and-phlex ([component here](https://github.com/jaredcwhite/test-bt-and-phlex/blob/main/src/_components/phlex_it.rb), [ERB template](https://github.com/jaredcwhite/test-bt-and-phlex/blob/main/src/index.md), [page with "inline phlex"](https://github.com/jaredcwhite/test-bt-and-phlex/blob/main/src/phlex-this.rb))
Demo: https://test-bt-and-phlex.onrender.com/

The rendering support is [added here](https://github.com/jaredcwhite/test-bt-and-phlex/blob/main/plugins/site_builder.rb) much like your Rails context mixin. I wonder if there's a way to DRY this up between the two.

I also experimented with an "inline phlex" helper which basically creates a component on the fly. Pretty neat-o in a page template. Probably not great in high-performance scenarios, but on a small static site it's definitely acceptable:
https://github.com/jaredcwhite/test-bt-and-phlex/blob/main/plugins/builders/inline_phlex.rb

lib/phlex/rails/context.rb
```diff
@@ -0,0 +1,33 @@
...
+        if @_rendering || @_rendering_block
+          _raw ApplicationControllerForPhlex.new.render_to_string(partial:, locals: kwargs)
```
alexandreruban, line 10: This trick about capturing the first positional argument as the partial name and passing the rest of the `kwargs` as locals seems risky, don't you think?

After some more digging inside the source code of Rails, maybe we could use this:

```suggestion
          _raw ApplicationControllerForPhlex.new.view_context.render(...)
```

That way we don't have to override the signature of the `render` method that is used inside Rails views. What do you think?

alexandreruban (approved):
Exciting!! 😊 

I just added one small comment about the signature of the `render` method, but otherwise it looks great!

lib/phlex/rails/context.rb
joeldrapper: Oh I like this idea.

PR #74, opened 2022-08-05
Simplify tag attributes

joeldrapper:
We don't need this special handling of CSS classes now that we don't support method chaining. #73 

Before:
```
                Page    18.000  i/100ms
                Page    182.792  (± 0.5%) i/s -      3.672k in  20.089634s
```

After:
```
                Page    22.000  i/100ms
                Page    227.200  (± 0.4%) i/s -      4.554k in  20.044099s
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,35 +161,6 @@ end
 
 This badge component takes a color argument and uses [Ecase](https://github.com/joeldrapper/ecase) to build a list of Tailwind classes.
 
-```ruby
-class BadgeComponent < Phlex::Component
-  VALID_COLORS = [:sky, :teal, :rose, :slate, :orange]
-
-  def initialize(color:)
-    super
-  end
-
-  def template(&)
-    span(class:, &)
-  end
-
-  private
-
-  def class
-    [:px_1, :py_1, :rounded_full, :text_sm] + colors
-  end
-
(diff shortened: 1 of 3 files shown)


joeldrapper:
This will make Phlex the fastest view layer in the [`view-layer-benchmarks`](https://github.com/ParamagicDev/view-layer-benchmarks). 🎉 

```
Comparison:
               phlex:     5550.7 i/s
      view_component:     5460.6 i/s - 1.02x  (± 0.00) slower
               cells:     3636.5 i/s - 1.53x  (± 0.00) slower
            partials:     2550.4 i/s - 2.18x  (± 0.00) slower
            dry_view:      354.9 i/s - 15.64x  (± 0.00) slower
```

alexandreruban (approved):
LGTM! 😄

janklimo:
@joeldrapper this PR removed the example but left the text above it which is somewhat confusing now that the example doesn't exist:

https://github.com/joeldrapper/phlex/blob/d5a21fe0db73eb3f7b403ba7a657f62bff0d4095/README.md?plain=1#L160-L164

joeldrapper:
Thanks! I’ll get rid of it. Will replace with a better example once we add new conditional classes.

PR #76, opened 2022-08-05
Render to children String buffer immediately

joeldrapper:
Instead of first building a tree of elements and then rendering that tree, we render each element to a string buffer immediately. This brings a significant performance improvement and makes way for a compiler method to do simple String appends from the compiled template method.

Before:
```
                Page    22.000  i/100ms
                Page    236.390  (± 1.3%) i/s -      1.188k in   5.026415s
```

After:
```
                Page    25.000  i/100ms
                Page    251.364  (± 0.0%) i/s -      1.275k in   5.072336s
```

### DIFF
--- a/lib/phlex/parent_node.rb
+++ b/lib/phlex/parent_node.rb
@@ -5,12 +5,11 @@ module Phlex
     include Callable
 
     def children
-      @_children ||= []
+      @_children ||= +""
     end
 
     def call(buffer = +"")
-      children.each { _1.call(buffer) }
-      buffer
+      buffer << children
     end
   end
 end
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -70,10 +70,6 @@ module Phlex
       @_target || self
     end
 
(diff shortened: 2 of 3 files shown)


lib/phlex/parent_node.rb
```diff
@@ -5,12 +5,11 @@ module ParentNode
     include Callable
 
     def children
-      @_children ||= []
+      @_children ||= +""
```
joeldrapper, line 8: I wonder if we should rename this `@_content` instead. 🤔

alexandreruban: Yes, I think it would make more sense to rename this now :)

alexandreruban (approved):
Nice!! :)

lib/phlex/parent_node.rb
joeldrapper: Can’t rename to "content" because it's taken by the `content(&)` method. I’m going to merge as is for now.

PR #77, opened 2022-08-08
Named Tag symbols

joeldrapper:
Tag symbols, like `<`, `>`, `</` and ` />` are now constant frozen strings. This is a very minor performance improvement but every little counts.


alexandreruban (approved):
LGTM! 💯

PR #78, opened 2022-08-08
Replace node objects with methods

joeldrapper:
Early on, I created various types of node objects: components, standard elements, void elements, text nodes, etc. This made it easy to read the code and build a picture in your head of what was going on. But all these nodes are object allocations — big ones, too. And they're not necessary.

This PR removes all node types apart from component and instead uses methods to concat on the buffer string. The component node is updated to be faster too be removing the step where it stores its own content buffer string before pushing it on the main buffer.

The result of this PR is a 44% performance improvement on our main benchmark.

Before:
```
                Page    25.000  i/100ms
                Page    249.084  (± 0.4%) i/s -      1.250k in   5.018461s
```

After:
```
                Page    36.000  i/100ms
                Page    359.743  (± 0.3%) i/s -      1.800k in   5.003611s
```


lib/phlex/block.rb
```diff
@@ -10,7 +10,7 @@ def initialize(context, &block)
     end
 
     def call(target, *args, **kwargs)
-      @context.render_block(target, *args, **kwargs, &@block)
+      @context._render_block(target, *args, **kwargs, &@block)
```
joeldrapper, line 13: Renamed this as it's a private method.

lib/phlex/component.rb
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
-require "digest"
```
joeldrapper, line 3: This was left over from the caching checksum stuff we removed #45.

lib/phlex/context.rb
```diff
@@ -2,16 +2,20 @@
 
 module Phlex
   module Context
+    def content(&)
```
joeldrapper, line 5: This is a DSL method so felt more at home here on Context.

lib/phlex/component.rb
```diff
@@ -62,25 +60,9 @@ def call(buffer = +"")
...
       @_rendered = true
 
+      @_target = buffer
       template(&@_content)
-      super
-    end
-
-    def target
-      @_target || self
-    end
```
joeldrapper, line 71: We always reference `@_target` directly to save a method call.

lib/phlex/context.rb
```diff
@@ -2,16 +2,20 @@
...
+    end
+
     def text(content)
-      Text.new(content).call(target.children)
+      @_target << CGI.escape_html(content)
     end
 
     def whitespace
-      _raw(" ")
+      @_target << " "
```
joeldrapper, line 14: Skip the `_raw` method call.

lib/phlex/context.rb
```diff
@@ -24,12 +28,21 @@ def component(component, *args, **kwargs, &block)
...
+        component.new(**kwargs) { text(args.first) }.call(@_target)
       else
-        component.new(*args, **kwargs, &block).call(target.children)
+        component.new(*args, **kwargs, &block).call(@_target)
       end
     end
 
+    def _render_block(new_target, ...)
+      old_target = @_target
+      @_target = new_target
```
alexandreruban, line 39: Do we still need those? Now `@_target` is always going to be the same buffer right?

alexandreruban (approved):
This looks really clear to me, and impressive performance gain!! 🤯

lib/phlex/context.rb
joeldrapper: You’re right, we don't need it.

PR #80, opened 2022-08-09
Cache tag attributes from their hash value

joeldrapper:
11.8% performance improvement on our benchmark.

Before:
```
                Page    35.000  i/100ms
                Page    355.176  (± 0.8%) i/s -      1.785k in   5.025994s
```

After:
```
                Page    39.000  i/100ms
                Page    397.184  (± 0.3%) i/s -      1.989k in   5.007804s
```

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -68,9 +68,24 @@ module Phlex
     end
 
     def _attributes(attributes)
+      if (cached = Phlex::ATTRIBUTE_CACHE[attributes.hash])
+        return @_target << cached
+      end
+
+      first_render = !self.class.rendered_at_least_once
+
+      buffer = first_render ? buffer = +"" : buffer = @_target
+
       attributes.transform_values! { CGI.escape_html(_1) }
       attributes[:href].sub!(/^\s*(javascript:)+/, "") if attributes[:href]
-      attributes.each { |k, v| @_target << Tag::SPACE << k.name << '="' << v << '"' }
+
+      attributes.each do |k, v|
+        buffer << Tag::SPACE << k.name << Tag::EQUALS_QUOTE << v << Tag::QUOTE
+      end
+
+      if first_render
+        @_target << Phlex::ATTRIBUTE_CACHE[attributes.hash] = buffer.freeze
+      end
     end
 
     Tag::STANDARD_ELEMENTS.each do |tag_name|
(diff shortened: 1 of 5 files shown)


joeldrapper:
One way to solve the memory leak might be to memoize if a component has been rendered before and don't do any new caching if it's been rendered once before. Either use the cache that's there or generate fresh attributes each time, but don't keep adding to the cache.

joeldrapper:
I fixed the memory leak by tracking whether a component class has been rendered before. Also added a spec to ensure nothing leaks.

lib/phlex/context.rb
```diff
@@ -44,7 +44,7 @@ def _standard_element(content = nil, _name: nil, **kwargs, &block)
       name = _name ||= __callee__.name
 
       @_target << Tag::LEFT << name
-      _attributes(kwargs)
+      _attributes(kwargs) if kwargs
```
grncdr, line 47: can `kwargs` ever be nil?

joeldrapper: Oh good point.

grncdr: still might be worth doing `if kwargs.any?` but I wouldn't be surprised if that was slower than falling through to the cache for an empty hash.

joeldrapper: This turned into a huge performance gain. See #81

lib/phlex/context.rb
```diff
@@ -68,9 +68,24 @@ def _void_element(**kwargs)
     end
 
     def _attributes(attributes)
+      if (cached = Phlex::ATTRIBUTE_CACHE[attributes.hash])
+        return @_target << cached
+      end
+
+      first_render = !self.class.rendered_at_least_once
```
alexandreruban, line 75: Maybe we could rename `self.class.rendered_at_least_once` into `self.class.first_render` to avoid the assignment to the variable and also have one single naming as we only use this here. What do you think?

alexandreruban (approved):
Nice!!

lib/phlex/context.rb
joeldrapper: The assignment to variable makes it faster because local variables are faster than multiple method calls. The variable is named positively on the class because the default value of `nil` is falsy. If `self.class.first_render` is `nil`, that's falsy but it should be truth.

PR #83, opened 2022-08-09
add install instructions when using Rails

davekaro:
Thanks @grncdr for the guidance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,15 @@ Or install it yourself as:
     $ gem install phlex
 
 
+### Using with Rails
+
+Update the Gemfile to require rails support
+
+```ruby
+gem 'phlex', require: 'phlex/rails'
+```
+
+or `require 'phlex/rails'` in application.rb
 
 ## License
 


grncdr:
As an alternative, what about adding something like the following to lib/phlex.rb

```rb
require 'phlex/rails' if defined?(Rails)
```

/cc @joeldrapper and @alexandreruban

joeldrapper:
@davekaro thanks so much for opening this PR. I really appreciate it! I think we'll probably go @grncdr’s route but I’ll keep this open until we make a decision on that.

davekaro:
From my perspective I would expect to add the gem and be done. So I favor the alternative.

joeldrapper:
👍 We're trying to keep some separation between Phlex and Rails because Rails isn't a dependency and Phlex can be used without it but I do like this detection idea.

davekaro:
Makes sense. What about Roda or Hanami users? Hard to say what support might be required without more research, but I think in both cases it would be nice to only add the gem in the Gemfile. Certainly better than N sections in the README for each web framework. Another data point, up until fairly recently view_component required a special require statement too. They deprecated that in https://github.com/github/view_component/pull/1162.

joeldrapper:
Closing this as it'll be covered by #91 or #82.

PR #87, opened 2022-08-11
Improve text content argument handling

joeldrapper:
Check if your initialiser handles positional arguments before assuming the first positional argument should be coerced into a block of text content.

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -5,32 +5,18 @@ module Phlex
     include Context
 
     module Overrides
-      module InitWithContent
-        def initialize(content = nil, *args, **kwargs, &block)
-          raise ArgumentError if block_given? && content
-
-          if block_given? && !block.binding.receiver.is_a?(Block)
-            block = Block.new(self, &block)
-          elsif content
-            block = Block.new(self) { text content }
-          end
-
-          @_content = block
-
-          super(*args, **kwargs)
+      def initialize(*args, **kwargs, &block)
+        if block_given? && !block.binding.receiver.is_a?(Block)
+          block = Block.new(self, &block)
         end
-      end
 
-      module InitWithoutContent
-        def initialize(*args, **kwargs, &block)
-          if block_given? && !block.binding.receiver.is_a?(Block)
-            block = Block.new(self, &block)
-          end
-
-          @_content = block
-
-          super(*args, **kwargs)
+        if !block_given? && args[0] && !method(__method__).super_method.parameters.to_h[:req]
+          content = args.delete_at(0)
+          block = Block.new(self) { text content }
         end
+
+        @_content = block
+        super(*args, **kwargs)
       end
 
       def template(...)
@@ -67,14 +53,7 @@ module Phlex
       end
 
       def inherited(child)
-        if child.instance_method(:initialize).parameters.to_h[:req]
-          child.prepend(Overrides::InitWithoutContent)
-        else
-          child.prepend(Overrides::InitWithContent)
-        end
-
         child.prepend(Overrides)
-
         super
       end
     end
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -10,7 +10,9 @@ module Phlex
           block = Block.new(self, &block)
         end
 
-        if !block_given? && args[0] && !method(__method__).super_method.parameters.to_h[:req]
+        super_method_params = method(__method__).super_method.parameters.to_h
+
+        if !block_given? && args[0] && !super_method_params[:req] && !super_method_params[:opt]
           content = args.delete_at(0)
           block = Block.new(self) { text content }
         end
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -5,14 +5,32 @@ module Phlex
     include Context
 
     module Overrides
-      def initialize(*args, **kwargs, &block)
-        if block_given? && !block.binding.receiver.is_a?(Block)
-          block = Block.new(self, &block)
+      module InitWithContent
+        def initialize(content = nil, *args, **kwargs, &block)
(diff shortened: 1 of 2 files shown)


lib/phlex/component.rb
```diff
@@ -49,7 +67,14 @@ def #{tag_name}(*args, **kwargs, &)
       end
 
       def inherited(child)
+        if child.instance_method(:initialize).parameters.to_h[:req]
```
joeldrapper, line 70: Sadly this doesn't work because the child hasn't defined its own `initialise` method by this point.

eregon:
IMHO this is too much magic (too hard to understand the behavior as it breaks encapsulation pretty drastically) and of course it's quite slow.
Also saw the `block.binding.receiver` above and that's also pretty slow and too magic IMHO.

If it's for fun, please go ahead, but if it's meant to be used project I would advise against getting any caller information and rather make things more explicit and less magic.

joeldrapper:
> Also saw the `block.binding.receiver` above and that's also pretty slow and too magic IMHO.

The project lives or dies on this, as far as I can tell. But Phlex is already about 3x faster than ActionView and 32% faster than ViewComponent.

eregon:
Did you measure the effect of this PR on perf? Would be interesting.

Re Proc#binding there are ideas floating around to remove it, because it causes unnecessary overhead for Ruby VMs.
But maybe it's not too bad perf-wise in this case.
From the code it seems you are using it to access the caller's self, right?

joeldrapper:
> From the code it seems you are using it to access the caller's self, right?

Yes, that's right. We need a to-procable object that lets us access the caller.

joeldrapper:
This PR seems to have no effect on our internal benchmark but it does negatively affect the the view-layer-benchmarks mentioned above.

```before
               phlex:     7047.8 i/s
      view_component:     5326.9 i/s - 1.32x  (± 0.00) slower
               cells:     3444.0 i/s - 2.05x  (± 0.00) slower
            partials:     2458.9 i/s - 2.87x  (± 0.00) slower
            dry_view:      345.9 i/s - 20.37x  (± 0.00) slower
```

After
```
               phlex:     6136.2 i/s
      view_component:     5379.7 i/s - 1.14x  (± 0.00) slower
               cells:     3494.1 i/s - 1.76x  (± 0.00) slower
            partials:     2468.5 i/s - 2.49x  (± 0.00) slower
            dry_view:      343.4 i/s - 17.87x  (± 0.00) slower
```

joeldrapper:
If checking whether `initialize` takes positional arguments is slow, perhaps we could just memoize that on the class.

> IMHO this is too much magic (too hard to understand the behavior as it breaks encapsulation pretty drastically) and of course it's quite slow.

Let me explain the intended behaviour here.

If I define a component that takes block contents

```ruby
class CardComponent
  def template(&)
    div class: "rounded drop-shadow p-5", &
  end
end
```

I should be able to pass that component either a block:
```ruby
component(CardComponent) { h1 "Hello world!" }
```

Or a string:
```ruby
component CardComponent, "Hello world!"
```

But if I define a component that explicitly takes a positional argument

```ruby
class NavItem
  def initialize(label, to:)
    @label = label
    @to = to
  end

  def template
    li { a @label, to: @to }
  end
end
```

Then passing a positional argument shouldn't be received as a block but rather delegated directly to the original initialiser.

```ruby
component NavItem, "Home", "/"
```

eregon:
Isn't it good enough to require `component(CardComponent) { "Hello world!" }` instead of `component CardComponent, "Hello world!"`? The block/Proc will be needed anyway, and then mixing initialize and template arguments is no longer a problem.

joeldrapper:
> The called_from = caller_locations(1,1)[0].label.to_sym in Overrides#template is another slow caller info access.

We could get rid of that override by renaming the method to something that doesn't conflict with `<template>` tags, or having a special case method for `<template>` tags.

Back to the issue of this PR, we could just forget the implicit first-argument content feature or only allow keyword arguments from component initialisers. Either is acceptable IMO though not ideal.

I am concerned by this though.

> Re Proc#binding there are ideas floating around to remove it, because it causes unnecessary overhead for Ruby VMs.

We need a to-procable that yields to the first caller and the only I can see to do that is to use our own block type that holds a reference to the context (`self`) from the first use.

eregon:
> We need a to-procable that yields to the first caller and the only I can see to do that is to use our own block type that holds a reference to the context (`self`) from the first use.

So you need the caller self but don't want to pass it explicitly, and just `self` is not what you want, right?
Potentially you could capture the caller self differently, e.g., with an explicit stack or save/restore pattern.
Maybe there should be `Proc#receiver` or so.

Why Proc#binding is problematic for optimization is:
```
def foo
  -> a, b { a + b }
end
```
Even such a lambda needs to capture the foo frame, because Proc#binding exposes it.
Without Proc#binding, and with a way to check at parse time for `Kernel#binding` and `*eval`, we wouldn't need to capture the foo frame here for the lambda. And even if the block captured variables we could just have a box to be able to write those captured variables but not need the full frame.

joeldrapper:
Yes, the methods that accept blocks check if the block is already a `Phlex::Block` by checking the `binding.receiver`. If not a `Phlex::Block`, they create a `Phlex::Block` which stores the original block and `self`.

```ruby
if block_given? && !block.binding.receiver.is_a?(Phlex::Block)
  block = Phlex::Block.new(self, &block)
end
```

Phlex blocks respond to `to_proc` so can be coerced with `&` and passed down many times as an anonymous block. When the block is yielded, that hits `Phlex::Block#call` which executes the original captured block against the original captured context.

```ruby
def call(*args, **kwargs)
  @context.instance_exec(*args, **kwargs, &@block)
end
```

`Proc#receiver` would be sufficient for our use-case. Or perhaps a native way to “instance_yield” a block.

PR #90, opened 2022-08-12
Remove template override

joeldrapper:
After the discussion with @eregon in #87, I think it'd be best to remove this override. I don’t think it’s worth the cost of performance and the extra complexity. Now, if you want a `<template>` tag, you need to use the `template_tag` method.

I also updated `Component#render` that was depending on this override unnecessarily. We actually don't need to respond to `render` to provide an HTML-safe render, as Rails will always call `render_in` if it's available. See #82 for more work on this where I’m trying to pass the view_context around so we can access the current request.

### Performance

Before
```
               phlex:     6919.5 i/s
      view_component:     5304.7 i/s - 1.30x  (± 0.00) slower
               cells:     3413.6 i/s - 2.03x  (± 0.00) slower
            partials:     2433.1 i/s - 2.84x  (± 0.00) slower
            dry_view:      337.7 i/s - 20.49x  (± 0.00) slower
```

After
```
               phlex:     8061.7 i/s
      view_component:     5290.0 i/s - 1.52x  (± 0.00) slower
               cells:     3442.0 i/s - 2.34x  (± 0.00) slower
            partials:     2428.9 i/s - 3.32x  (± 0.00) slower
            dry_view:      338.0 i/s - 23.85x  (± 0.00) slower
```

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -14,21 +14,6 @@ module Phlex
 
         super(*args, **kwargs)
       end
-
-      def template(...)
-        called_from = caller_locations(1,1)[0].label.to_sym
-
-        # If we're rendering a template and not being called from
-        # another override in the ancestry, we render the template tag.
-
-        if @_rendering && (called_from != __method__)
-          _template_tag(...)
-        else
-          @_rendering = true
-          super
-          @_rendering = false
-        end
-      end
     end
 
     class << self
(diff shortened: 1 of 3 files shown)


jaredcwhite:
That seems like a reasonable compromise. The frequency with which people will be adding literal template elements into their components is likely minimal. And look at that speed gain! 😲

alexandreruban (approved):

joeldrapper:
Alright, no strong objections so far. I’m going to ship this but of course we can always re-visit if we come up with a better way forward.

PR #93, opened 2022-08-12
Detect unsafe attribute names

### LINKED ISSUE #92: Safe attribute keys
To add some extra protection against unsafe content, we should remove certain characters from attribute keys. We don’t need to do a full HTML escape because we don't need to preserve anything. Should probably gsub the following `<`, `>`, `"`, `'`, `&`.

It's already quite difficult to shoot yourself in the foot with this. You’d need to coerce a hash with unsafe keys into the arguments for a tag, e.g. `div **unsafe_hash`.

joeldrapper:
Closes #92 

This didn't have any effect on our benchmarks because of the attribute caching so I added a unique attribute to the benchmark. With that, there appears to be a no significant performance impact when the attributes can't be cached. My benchmarks actually showed it was faster, but within error margins.

Before
```
                Page    40.000  i/100ms
                Page    400.399  (± 0.2%) i/s -      2.040k in   5.094980s
```

After
```
                Page    40.000  i/100ms
                Page    400.561  (± 0.2%) i/s -      2.040k in   5.092894s
```

### DIFF
--- a/lib/phlex.rb
+++ b/lib/phlex.rb
@@ -7,6 +7,9 @@ loader = Zeitwerk::Loader.for_gem
 loader.setup
 
 module Phlex
+  Error = Module.new
+  ArgumentError = Class.new(ArgumentError) { include Error }
+
   extend self
 
   ATTRIBUTE_CACHE = {}
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -76,7 +76,17 @@ module Phlex
 
       buffer = first_render ? buffer = +"" : buffer = @_target
 
+      attributes.each_key do |key|
+        if key.match? /[<>&"']/
+          raise ArgumentError, <<~MESSAGE
+            Unsafe attribute name detected.
+            Attributes names shouldn't contain `<`, `>`, `&`, `"` or `'`.
+          MESSAGE
+        end
+      end
+
       attributes.transform_values! { CGI.escape_html(_1) }
+
       attributes[:href].sub!(/^\s*(javascript:)+/, "") if attributes[:href]
 
       attributes.each do |k, v|
--- a/spec/phlex/component_spec.rb
+++ b/spec/phlex/component_spec.rb
@@ -123,6 +123,26 @@ RSpec.describe Phlex::Component do
     end
   end
 
+  describe "with unsafe attributes" do
+    let :component do
+      Class.new Phlex::Component do
+        def template
+          article **attributes
+        end
+
+        def attributes
+          {
+            %{"><script type="text/javascript">alert(1);</script>} => "test"
+          }
+        end
+      end
+    end
+
+    it "escapes the content" do
+      expect { output }.to raise_error ArgumentError
+    end
+  end
+
   describe "with javascript link protocol in hypertext reference" do
     let :component do
       Class.new Phlex::Component do
--- a/examples/page.rb
+++ b/examples/page.rb
@@ -7,7 +7,7 @@ module Example
         5.times do
           div do
             10.times do
-              a "Test", href: "something"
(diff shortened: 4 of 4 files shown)


lib/phlex/context.rb
```diff
@@ -76,7 +76,17 @@ def _attributes(attributes)
 
       buffer = first_render ? buffer = +"" : buffer = @_target
 
+      attributes.keys.each do |key|
+        if key.match /[<>&"']/
+          raise ArgumentError, <<~MESSAGE
```
joeldrapper, line 81: Rather than correct unsafe attribute names, I decided to raise because checking for their presence is much faster than gsubbing, even if they’re not there. I think this is a good approach anyway as you should never expect to have unsafe attribute names and it's better to be noisy about it.

lib/phlex.rb
```diff
@@ -7,6 +7,9 @@
 loader.setup
 
 module Phlex
+  Error = Module.new
+  ArgumentError = Class.new(ArgumentError) { include Error }
```
joeldrapper, line 11: This trick allows us to have a `Phlex::ArgumentError` that `is_a?(ArgumentError)` and `is_a?(Phlex::Error)`.

spec/phlex/component_spec.rb
```diff
@@ -123,6 +123,26 @@ def template
     end
   end
 
+  describe "with unsafe attributes" do
+    let :component do
+      Class.new Phlex::Component do
+        def template
+          article **attributes
```
joeldrapper, line 130: It’s not likely this would ever happen, but the point is it could and we should check.

lib/phlex/context.rb
```diff
@@ -76,7 +76,17 @@ def _attributes(attributes)
 
       buffer = first_render ? buffer = +"" : buffer = @_target
 
+      attributes.keys.each do |key|
+        if key.match /[<>&"']/
```
joeldrapper, line 80: I benchmarked `match` against `in` and `match` is faster. I also benchmarked joining the keys first and matching on the joined String with varying numbers of symbols. Joining first much slower.

gmammadova, line 80: Not sure if Ruby interpreter optimizes this out automatically, but if not, would initializing the regex object outside of the loop help to squeeze out a few more CPU cycles?

ElvinEfendi, line 80: Not sure if Ruby interpreter optimizes this out automatically, but if not, would initializing the regex object outside of the loop help to squeeze out a few more CPU cycles?

bdewater: Looks like `match?` can be used here, which should be the fastest: https://github.com/fastruby/fast-ruby#regexp-vs-stringmatch-vs-string-vs-stringmatch-code-

joeldrapper: Thanks. That's a great idea.

byroot: Regexp literals are interned, so no it would actually be worse because you'd have to lookup a local variable or something like that.

lib/phlex/context.rb
```diff
@@ -76,7 +76,17 @@ def _attributes(attributes)
 
       buffer = first_render ? buffer = +"" : buffer = @_target
 
+      attributes.keys.each do |key|
```
byroot, line 79: if you want ROFLscale
```suggestion
      attributes.each_key do |key|
```

Saves an intermediate array.

joeldrapper: Why not‽

lib/phlex/context.rb
joeldrapper: Thanks for this.

PR #101, opened 2022-08-15
Boolean attributes

joeldrapper:
Support boolean HTML attributes, e.g. `input type: "radio", checked: true`.

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -80,12 +80,21 @@ module Phlex
         end
       end
 
-      attributes.transform_values! { CGI.escape_html(_1) }
+      attributes.transform_values! do |value|
+        next value if (value == true || value == false)
+        CGI.escape_html(value)
+      end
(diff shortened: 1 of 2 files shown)


lib/phlex/context.rb
```diff
@@ -80,12 +80,21 @@ def _attributes(attributes)
...
+        CGI.escape_html(value)
+      end
 
       attributes[:href].sub!(/^\s*(javascript:)+/, "") if attributes[:href]
 
       attributes.each do |k, v|
-        buffer << Tag::SPACE << k.name << Tag::EQUALS_QUOTE << v << Tag::QUOTE
+        next unless v
+
+        if v == true
```
alexandreruban, line 93: @joeldrapper shouldn't we check for the `false` value here as well?

joeldrapper: It's checked on line 91.

PR #103, opened 2022-08-16
Add pointer cursor to tab headers to make them look clickable

djfpaagman:
Was having a look at the docs and noticed that the tab headers didn't have the pointer cursor that makes them look clickable like any other link.

Small fix for that in this PR as I wanted to see if I can contribute a bit to this project :)

I think the tabs are still a bit weird, as they get focus when you click one and you have to click again to unfocus before you can click something else, but not sure how to fix that properly without rewriting the whole component.


alexandreruban (approved):
Makes sense!

PR #105, opened 2022-08-16
Override `.new` instead of `#initialize`

joeldrapper:
If we override `.new`, we don't need to prepend a module to override `#initialize`. This seems to be faster and it also keeps the ancestry nice and clean, especially when subclassing a component, since every time a component is subclassed, a new override would have been prepended on it.

Before:
```
                Page    70.000  i/100ms
                Page    696.288  (± 1.0%) i/s -      3.500k in   5.027149s
```

After:
```
                Page    71.000  i/100ms
                Page    701.157  (± 1.0%) i/s -      3.550k in   5.063536s
```

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -4,17 +4,20 @@ module Phlex
   class Component
     include Context, Renderable
 
-    module Overrides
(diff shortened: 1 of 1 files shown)


alexandreruban (approved):
Great idea!

PR #109, opened 2022-08-16
Dasherize HTML tag attributes

joeldrapper:
I think it makes sense to automatically dasherize HTML tag attributes, since this is the norm and you can't easily use dashes in symbols.


alexandreruban (approved):
Nice!

PR #114, opened 2022-08-18
Support Ruby 2.7

joeldrapper:
Looks like we can support Ruby 2.7 with just one patch to Symbol to make it alias `name` to `to_s.freeze`. Using `name` instead of `to_s` is faster in newer Rubies.

With these changes, specs pass in Ruby 2.7.6 without any warnings.

### DIFF
--- a/lib/phlex.rb
+++ b/lib/phlex.rb
@@ -7,6 +7,16 @@ loader = Zeitwerk::Loader.for_gem
 loader.ignore("#{__dir__}/generators")
 loader.setup
 
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("3.0")
+  unless Symbol.instance_methods.include?(:name)
+    class Symbol
+      def name
+        to_s.freeze
+      end
+    end
+  end
+end
+
 module Phlex
   Error = Module.new
   ArgumentError = Class.new(ArgumentError) { include Error }
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -14,6 +14,7 @@ module Phlex
           @_view_context = _view_context
           @_parent = _parent
           @_content = block
+          @_rendered = false
         end
 
         component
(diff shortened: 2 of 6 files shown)


lib/phlex/component.rb
```diff
@@ -14,6 +14,7 @@ def new(*args, _view_context: nil, _parent: nil, **kwargs, &block)
           @_view_context = _view_context
           @_parent = _parent
           @_content = block
+          @_rendered = false
```
joeldrapper, line 17: This fixes a Ruby 2.7 warning about using an uninitialised instance variable.

spec/phlex/component_spec.rb
```diff
@@ -148,7 +148,7 @@ def template
     let :component do
       Class.new Phlex::Component do
         def template
-          article **attributes
+          article(**attributes)
```
joeldrapper, line 151: This fixes an ambiguity warning in Ruby 2.7.

lib/phlex/context.rb
```diff
@@ -76,7 +76,7 @@ def _attributes(attributes)
       attributes.each do |k, v|
         next unless v
 
-        if k.match? /[<>&"']/
+        if k.match?(/[<>&"']/)
```
joeldrapper, line 79: This fixes an ambiguity warning in Ruby 2.7.

lib/phlex.rb
```diff
@@ -7,6 +7,16 @@
 loader.ignore("#{__dir__}/generators")
 loader.setup
 
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("3.0")
+  unless Symbol.instance_methods.include?(:name)
+    class Symbol
+      def name
+        to_s.freeze
```
joeldrapper, line 14: This is slightly slower than if we just aliased `name` to `to_s`, but it’s a more faithful counterpart to the `Symbol#name` method in Ruby >= 3.0.

PR #121, opened 2022-08-22
Block event attributes

### LINKED ISSUE #119: Block event attributes
All the event attributes should be blocked by default. I found this list but I’m not sure if there are others. https://www.w3schools.com/tags/ref_eventattributes.asp

joeldrapper:
Closes #119

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -72,11 +72,8 @@ module Phlex
       attributes.each do |k, v|
         next unless v
 
-        if k.match?(/[<>&"']/)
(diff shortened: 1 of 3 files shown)


joeldrapper:
I tried implementing this with a specific disallow list from [here](https://www.w3schools.com/tags/ref_eventattributes.asp) but that ended up being really slow. This approach searches for `on[a-z]` at the start of the attribute name disallows that. Unfortunately it could have unwanted side effects for attributes like `only`, but according to the HTML5 spec, that should really be `data-only`. 🤷

PR #122, opened 2022-08-23
Allow unsafe attributes in a `without_attribute_safety` block

joeldrapper:
I don’t know if we should ship this feature without a strong use-case for it. You shouldn’t ever need to use HTML event attributes or `javascript:` links and these should ideally be blocked in your CSP anyway.


joeldrapper:
Closing this as I’d like to use an alternative approach where each attribute can be wrapped in a method that declares it’s safe.

PR #123, opened 2022-08-24
Add missing `link`, `style`, `meta` elements

jaredcwhite:
Very useful in `<head>` scenarios and possibly other places (`link` and `style` can go in `body` as well).

Q: any tests needed to update?


joeldrapper:
Nope, we don't test every tag in the configuration, just a few specific examples.

PR #124, opened 2022-08-25
All-caps tag methods

joeldrapper:
To avoid conflicts with other method names such as [Kernel#p](https://rubyapi.org/3.1/o/kernel#method-i-p) and `Phlex::Component#template` we could make all the tag methods uppercase.

I don't think it looks or feels as nice as it did before, but ultimately I do think this could lead to a better developer experience. It's entirely reasonable, for example, to define a `title` method and then try to output that in a `<title>` element.

Before, this could have resulted in an infinite loop. Now, it just works.

```diff
def template
-  title(title)
+  TITLE(title)
end

def title
  ...
end
```

One downside is all-caps method calls without arguments will be treated as constant references unless they have empty parens.

Some other options:
1. Live with the potential conflicts — it's very unlikely anyone will need to use Kernel#p from a Phlex component and it's not difficult to avoid new conflicts.
2. Underscore prefix, e.g. `_p`
3. Underscore suffix, e.g. `p_`
4. `_tag` suffix, e.g. `p_tag`

### DIFF
--- a/docs/components/callout.rb
+++ b/docs/components/callout.rb
@@ -1,7 +1,7 @@
 module Components
   class Callout < Phlex::Component
     def template(&)
-      div(class: "rounded bg-orange-50 text-sm p-5 border border-orange-100", &)
+      DIV(class: "rounded bg-orange-50 text-sm p-5 border border-orange-100", &)
     end
   end
 end
--- a/docs/components/code_block.rb
+++ b/docs/components/code_block.rb
@@ -8,7 +8,7 @@ module Components
     end
 
     def template
-      pre(class: "highlight p-5 whitespace-pre-wrap bg-stone-50") {
+      PRE(class: "highlight p-5 whitespace-pre-wrap bg-stone-50") {
         _raw FORMATTER.format(
           lexer.lex(@code)
         )
--- a/docs/components/heading.rb
+++ b/docs/components/heading.rb
@@ -1,7 +1,7 @@
 module Components
   class Heading < Phlex::Component
     def template(&)
-      h2(class: "text-xl font-bold mt-10 mb-5", &)
+      H2(class: "text-xl font-bold mt-10 mb-5", &)
     end
   end
 end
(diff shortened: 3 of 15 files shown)


joeldrapper:
I feel quite strongly that we should not do this. The benefits of the cleaner syntax and not having to remember to use empty parens outweigh the relatively low risk of conflicts. Each component is isolated, so conflicts can't leak between components and will always be pretty easy to spot.

diegotoral:
I have mixed feelings about this change. The code with uppercase methods does not look like real ruby code for me.

jaredcwhite:
Not a fan of the caps TBH, doesn't pass the smell test.

> 1. Live with the potential conflicts — it's very unlikely anyone will need to use Kernel#p from a Phlex component and it's not difficult to avoid new conflicts.

All around this feels like the best compromise amidst various competing concerns.

> 2. Underscore prefix, e.g. `_p`

Anything prefixed with an underscore feels like a private-ish variable to me, so that doesn't look right at first glance.

> 3. Underscore suffix, e.g. `p_`

This would be my preference if not going with the standard tag names.

> 4. `_tag` suffix, e.g. `p_tag`

Too much extra typing IMHO.

RolandStuder:
I really dislike this suggestion, looks like you are using constants as methods, then a super nice ruby structure becomes this weird thing:

```ruby
      DIV(class: "rounded bg-orange-50 text-sm p-5 border border-orange-100", &)
```

That just does not look like ruby.

I think this all stemmed from the kernel having the `p` method. I used the `p` tag in my experiments without any second thought that it could mean something else. The context makes it sufficiently clear in my opinion, except maybe if you extract methods, so when you have something like:

```ruby
class MyComponent < Phlex::Component
  def template
    div {
      explanation
    }
  end

  def explanation
     p "I want to be seen as a tag not a Kernel.p"
  end
end
```

I would currently lean towards actually just overwriting the `p` method in your components and use it for the `p`-tag. To me that feels natural. When would you ever need a `Kernel.p` in a component, except for some debugging, where you should quickly realize, that this is unsurprisingly not using `Kernel.p`.

joeldrapper:
Thanks for all the feedback. It looks like we're all in agreement so I’m closing this issue.

PR #125, opened 2022-08-27
Table component

joeldrapper:
First-party `Phlex::Table` component allows you to build HTML tables with a DSL. First define the table:

```ruby
class Articles::Table < Phlex::Table
  property "Title" do |article|
    a article.title, href: articles_path(article)
  end

  property "Publication date", &:published_at
  property "Comments", &:comments_count
end
```

Then pass it a list of resources:

```ruby
render Articles::Table.new(collection: @articles)
```

Or a single resource to render a fragment:
```ruby
render Articles::Table.new(item: @article)
```

### DIFF
--- a/lib/phlex/table.rb
+++ b/lib/phlex/table.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module Phlex
+	class Table < Phlex::Collection
+		class << self
+			def property(header, &block)
+				if header.is_a?(String)
+					header_text = header
+					header = -> { text header_text }
+				end
+
+				properties << {
+					header: header,
+					body: block
+				}
+			end
+
+			def properties
+				@properties ||= []
+			end
+		end
+
+		def collection_template
+			table do
+				thead do
+					tr do
+						self.class.properties.each do |property|
+							th scope: "col" do
+								instance_exec(&property[:header])
+							end
+						end
+					end
+				end
+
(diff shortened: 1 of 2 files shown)


lib/phlex/table.rb
```diff
@@ -0,0 +1,97 @@
...
+      th(*args, **kwargs, &block)
+    end
+
+    def row(*args, **kwargs, &block)
+      tr(*args, **kwargs, &block)
+    end
+
+    def cell(*args, **kwargs, &block)
+      td(*args, **kwargs, &block)
+    end
```
joeldrapper, line 57: The default implementations for these could be faster alias methods if we changed the way we define tag methods.

lib/phlex/context.rb
```diff
@@ -7,11 +7,17 @@
...
+      unchanged = (original_size == @_target.size)
 
-      text(output) if unchanged && output.is_a?(String)
+      if unchanged
+        if output.is_a?(String)
+          text(output)
+        elsif output.respond_to?(:to_s)
+          text(output.to_s)
+        end
+      end
```
joeldrapper, line 20: Support to-stringables without unnecessarily calling `to_s` on strings.

joeldrapper:
I’m shipping this undocumented and untested a an experimental feature, with the intention of documenting and testing it later.

PR #126, opened 2022-08-29
Define tag methods with macros

joeldrapper:
Use macros to define the tag methods instead of `alias_method` with `__callee__`. This enables us to preemptively generate the entire opening and closing strings so there's less runtime concatenation. 

Before
```
                Page    71.000  i/100ms
                Page    714.729  (± 0.3%) i/s -      3.621k in   5.066311s
```

After
```
                Page    90.000  i/100ms
                Page    905.819  (± 0.3%) i/s -      4.590k in   5.067280s
```

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -6,6 +6,8 @@ end
 
 module Phlex
   module Context
+    extend Tag
+
     def content(&block)
       original_bytesize = @_target.size
       output = yield if block_given?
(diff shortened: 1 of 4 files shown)


lib/phlex/context.rb
```diff
@@ -19,7 +25,7 @@ def text(content)
     end
 
     def whitespace
-      @_target << " "
+      @_target << Tag::SPACE
```
joeldrapper, line 28: This saves one object allocation per boot, lol.

spec/phlex/component_spec.rb
```diff
@@ -387,7 +387,8 @@ def status_emoji
     describe "with custom elements" do
       let :component do
         Class.new Phlex::Component do
-          register_element :trix_editor, :trix_toolbar
+          register_element :trix_editor
+          register_element :trix_toolbar
```
joeldrapper, line 391: Not accepting a splat means we can accept a `tag:` override in case we need the method name to be different from the tag it generates.

lib/phlex/tag.rb
```diff
@@ -1,133 +1,58 @@
...
 module Phlex
   module Tag
     DASH = "-"
     SPACE = " "
     UNDERSCORE = "_"
     NAMESPACE_DELINEATOR = "::"
 
-    LEFT = "<"
     RIGHT = ">"
-    CLOSE_LEFT = "</"
```
joeldrapper, line 12: Removing these since we don't use them anymore.

PR #132, opened 2022-08-31
Initial rails form component

joeldrapper:
This component is meant to be the Phlex counterpart to `form_with`, but the goal is not compatibility with the `form_with` API. Some parts, e.g `options_for_select` are not worth copying.


joeldrapper:
I’m closing this as I think it was going in the wrong direction. Forms should themselves be classes that take the object(s) they represent as arguments. I think copying form_with was a bad idea.

PR #145, opened 2022-09-05
Flatten Hash attributes

joeldrapper:
This will allow you to provide Hash values for tag attributes, e.g. this

```ruby
div data: { value: 1 }
```

...would produce this:

```html
<div data-value="1"></div>
```

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -73,36 +73,51 @@ module Phlex
       @_target << content
     end
 
-    def _attributes(attributes)
-      first_render = !self.class.rendered_at_least_once
-
-      buffer = first_render ? buffer = +"" : buffer = @_target
-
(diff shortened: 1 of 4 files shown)


lib/phlex/component.rb
```diff
@@ -96,6 +96,8 @@ def _attributes(attributes)
           buffer << " " << k.name.tr("_", "-") << '="' << CGI.escape_html(v) << '"'
         when Symbol
           buffer << " " << k.name.tr("_", "-") << '="' << CGI.escape_html(v.name) << '"'
+        when Hash
+          _attributes(v.transform_keys { "#{k}-#{_1}".to_sym })
```
joeldrapper, line 100: Some of the stuff in `_attributes` doesn't need to be repeated for each nested attribute. We should probably extract another method, though I’m not sure what to call it.

PR #157, opened 2022-09-13
Initial Compiler

joeldrapper:
Initial setup for compilation. No compilation is happening yet, but it's available for value calls.


joeldrapper:
This doesn't do anything yet, but I want to merge it because it touches a lot of files and is likely to pick up conflicts.

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest', 'macos-latest']
-        ruby-version: ['2.7', '3.0', '3.1', 'head', 'truffleruby-22.2', 'truffleruby-head']
```
eregon, line 12: Was this intentional?
It seemed green the day before: https://github.com/joeldrapper/phlex/actions/runs/3128985230/jobs/5077506378

joeldrapper: Yes, unfortunately TruffleRuby doesn't support pattern matching, so we wouldn't be able to use SyntaxTree for the compiler. Kevin [mentioned](https://twitter.com/nirvdrum/status/1570394894622941185) there is ongoing work to support pattern matching, so I decided to continue working on the SyntaxTree compiler with the intention of adding TruffleRuby support back in later.

eregon: I see, thank you for explaining. One way could be using RubyNext in syntax_tree until TruffleRuby has full pattern matching support: https://github.com/ruby-syntax-tree/syntax_tree/issues/168

PR #162, opened 2022-09-16
Add SVG and PATH tags

lucianghinda:
This PR aims to add support for [`svg`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg) and [`path`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/path) elements.


joeldrapper:
Thanks @lucianghinda!

PR #165, opened 2022-09-18
Implicitly convert component content to string

### LINKED ISSUE #163: Implicit conversion of text content to string

marcoroth:
This Pull Requests implicitly converts the given content to a string so you, as a user of the library, don't have to convert the content upfront before passing it to the `text` or any `html` method.

Resolves #163

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -47,15 +47,20 @@ module Phlex
       return unless block_given?
 
       original_length = @_target.length
-      output = yield(self) if block_given?
+      output = yield(self)
       unchanged = (original_length == @_target.length)
 
-      text(output) if unchanged && output.is_a?(String)
+      text(output) if unchanged
       nil
     end
 
     def text(content)
-      @_target << CGI.escape_html(content)
+      @_target << case content
+      when String then CGI.escape_html(content)
+      when Symbol then CGI.escape_html(content.name)
+      else CGI.escape_html(content.to_s)
+      end
+
       nil
     end
 
--- a/lib/phlex/html.rb
+++ b/lib/phlex/html.rb
@@ -31,7 +31,14 @@ module Phlex
             end
           else
             if content
-              @_target << "<#{tag}>" << CGI.escape_html(content) << "</#{tag}>"
+              case content
+              when String
+                @_target << "<#{tag}>" << CGI.escape_html(content) << "</#{tag}>"
+              when Symbol
+                @_target << "<#{tag}>" << CGI.escape_html(content.name) << "</#{tag}>"
+              else
+                @_target << "<#{tag}>" << CGI.escape_html(content.to_s) << "</#{tag}>"
+              end
             elsif block_given?
               @_target << "<#{tag}>"
               content(&block)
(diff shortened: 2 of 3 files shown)


lib/phlex/component.rb
```diff
@@ -50,12 +50,12 @@ def content(&block)
       output = yield(self) if block_given?
```
joeldrapper, line 50: I know this is unrelated, but this conditional isn't necessary. Could we remove it in this PR?

```suggestion
      output = yield(self)
```

lib/phlex/component.rb
```diff
@@ -50,12 +50,12 @@ def content(&block)
...
       unchanged = (original_length == @_target.length)
 
-      text(output) if unchanged && output.is_a?(String)
+      text(output) if unchanged
       nil
     end
 
     def text(content)
-      @_target << CGI.escape_html(content)
+      @_target << CGI.escape_html(content.to_s)
```
joeldrapper, line 58:
```suggestion
      case content
      when String
        @_target << CGI.escape_html(content)
      when Symbol
        @_target << CGI.escape_html(content.name)
      else
        @_target << CGI.escape_html(content.to_s)
      end
```

lib/phlex/component.rb
```diff
@@ -70,7 +70,7 @@ def doctype
     end
 
     def raw(content)
-      @_target << content
+      @_target << content.to_s
```
joeldrapper, line 73: I don’t think we should implicitly convert raw content to a string. Raw output should be really explicit.

lib/phlex/html.rb
```diff
@@ -31,7 +31,7 @@ def #{element}(content = nil, **attributes, &block)
             end
           else
             if content
-              @_target << "<#{tag}>" << CGI.escape_html(content) << "</#{tag}>"
+              @_target << "<#{tag}>" << CGI.escape_html(content.to_s) << "</#{tag}>"
```
joeldrapper, line 34:
```suggestion
              case content
              when String
                @_target << "<#{tag}>" << CGI.escape_html(content) << "</#{tag}>"
              when Symbol
                @_target << "<#{tag}>" << CGI.escape_html(content.name) << "</#{tag}>"
              else
                @_target << "<#{tag}>" << CGI.escape_html(content.to_s) << "</#{tag}>"
              end
```

joeldrapper (changes requested):
Hey @marcoroth, thanks for the PR!

There’s a very small cost to calling `#to_s` on a String so I think we should avoid this with a case. We can also use `#name` to get the frozen string for Symbols which is faster than `#to_s`.

I also don't think we should implicitly convert raw content.

marcoroth:
Sweet, thanks for the review! I updated the branch with the `case` statements and removed the implicit conversion for `raw`.

joeldrapper (approved):

joeldrapper:
I'm not sure why Rubocop is complaining but it's more likely Rubocop is wrong than your code it. I'll take a look at it and see if the configuration needs changing.

marcoroth:
My bad, I amended the commit but forget to include the patch that fixed the RuboCop offense

PR #169, opened 2022-09-18
Add SOURCE tag

bb:
Add SOURCE tag for usage with picture element.


joeldrapper:
Thanks for opening this! Until this is released, you can use `register_void_element :source` from an `ApplicationComponent` superclass.

bb:
Thanks Joel!

PR #174, opened 2022-09-22
Spaces -> Tabs

joeldrapper:
Switching from spaces to tabs for accessibility. This code should be accessible to everyone.

To make this easier, I’ve updated the RuboCop config, `rubocop -A` should autocorrect any spaces to tabs. Additionally, I’ve added a `.editorconfig` with 2-width tabs by default.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*]
+indent_style = tab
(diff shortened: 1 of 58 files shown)


cmer:
That is a bold move, my friend.

PR #175, opened 2022-09-22
Add `comment` to support HTML comments

marcoroth:
Sometimes it can also be useful to add comments to your generated HTML. 

That's why this Pull Request introduces a `comment` method to `Phlex::Component` so that you can create HTML comments using the Phlex DSL.

### DIFF
--- a/lib/phlex/component.rb
+++ b/lib/phlex/component.rb
@@ -70,7 +70,7 @@ module Phlex
 		end
 
 		def comment(content = "")
-			@_target << "<!-- " << content.to_s << " -->"
+			@_target << "<!-- " << CGI.escape_html(content.to_s) << " -->"
 			nil
 		end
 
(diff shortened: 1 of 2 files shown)


marcoroth:
I've got two questions about the implementation: 

* Should the `comment` also append a newline?
* Do you prefer `@_target << "<!-- " << content.to_s << " -->"` over `@_target << "<!-- #{content} -->"`?

joeldrapper:
I really like this idea. I also think it would be useful to output comments around components in development since there's no HTML source maps standard.

We should probably html-escape comments just in case, and let the compiler optimise that away.

marcoroth:
> I really like this idea. I also think it would be useful to output comments around components in development since there's no HTML source maps standard.

Agreed! I'm wondering if we could somehow hook into `config.action_view.annotate_rendered_view_with_filenames`.

Also added the escape for the content 👍🏼

joeldrapper:
> Agreed! I'm wondering if we could somehow hook into `config.action_view.annotate_rendered_view_with_filenames`.

Yeah that's probably a good idea.

PR #177, opened 2022-09-23
fix: autoloading of rails engines

ParamagicDev:
This more closely follows the Rails convention for Engines. 

The only thing I'm not sure about is what happens if the Rails engine is loaded when the user doesn't have Rails so I added a "LoadError" check.

### DIFF
--- a/lib/phlex.rb
+++ b/lib/phlex.rb
@@ -25,10 +25,9 @@ module Phlex
 	end
 end
 
-
 begin
-  require "rails"
-  require "phlex/engine"
+	require "rails"
+	require "phlex/engine"
 rescue LoadError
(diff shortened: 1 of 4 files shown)


lib/phlex.rb
```diff
@@ -25,6 +25,9 @@ def configure
 	end
 end
 
-if defined?(Rails::Engine)
+begin
 	require "rails"
+	require "phlex/engine"
+rescue LoadError
+	# Rails isnt in this env, dont load the engine.
```
alexandreruban, line 32:
```suggestion
	# Rails isn't in this env, don't load the engine.
```

alexandreruban (approved):
LGTM!

PR #186, opened 2022-09-26
Add integration tests for ViewComponent

nikolalsvk:
Closes https://github.com/joeldrapper/phlex/issues/161

I removed `renderable.is_a?(Class)` in [renderable.rb](https://github.com/joeldrapper/phlex/blob/2cb6133035cc1bd8433f0931755321143e3ca6b4/lib/phlex/renderable.rb#L12) and tests fail without it

Does this make sense?


fixtures/dummy/app/components/comment.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <span>
+    <%= @body %>
+  </span>
+
+  <%= content %>
+
+  <%= render Views::Card.new do |a| %>
+    <p>Start Card A</p>
+    <%= a.title "Hello from A" %>
+  <% end %>
```
joeldrapper, line 14: I like this, but can we put the `@name` and `@body` instance variables in the card to be sure they work in the block?

fixtures/dummy/app/components/comment.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class Comment < ViewComponent::Base
```
joeldrapper, line 3: We should rename this `CommentComponent`.

test/phlex/integration/view_component/render.rb
```diff
@@ -0,0 +1,23 @@
...
+require "test_helper"
+
+describe Phlex::Component do
+	with "ViewComponent renders a Phlex component" do
+		let(:output) { CommentsController.render "index" }
+
+		it "renders comments" do
+			expect(output).to be ==
+										%(<div>\n  <span>\n    Matz\n  </span>\n  <span>\n    hey, folks\n  </span>\n\n  \n  Hello, World from a ViewComponent!\n\n\n  \n    <p>Start Card A</p>\n    \n<article class=\"drop-shadow p-5 rounded\"><h3 class=\"font-bold\">Hello from A</h3></article></div>\n)
+		end
```
joeldrapper, line 12:
```suggestion
		it "renders comments" do
			expect(output).to be ==
				%(<div>\n  <span>\n    Matz\n  </span>\n  <span>\n    hey, folks\n  </span>\n\n  \n  Hello, World from a ViewComponent!\n\n\n  \n    <p>Start Card A</p>\n    \n<article class=\"drop-shadow p-5 rounded\"><h3 class=\"font-bold\">Hello from A</h3></article></div>\n)
		end
```

test/phlex/integration/view_component/render.rb
```diff
@@ -0,0 +1,23 @@
...
+	end
+
+	with "Phlex component renders a ViewComponent" do
+		let(:output) { CommentsController.render "show" }
+
+		it "renders comments" do
+			expect(output).to be ==
+										%(<div>\n  <span>\n    _why\n  </span>\n  <span>\n    I&#39;m back\n  </span>\n\n  \n\n  \n    <p>Start Card A</p>\n    \n<article class=\"drop-shadow p-5 rounded\"><h3 class=\"font-bold\">Hello from A</h3></article></div>\n\n)
+		end
```
joeldrapper, line 21:
```suggestion
		it "renders comments" do
			expect(output).to be ==
				%(<div>\n  <span>\n    _why\n  </span>\n  <span>\n    I&#39;m back\n  </span>\n\n  \n\n  \n    <p>Start Card A</p>\n    \n<article class=\"drop-shadow p-5 rounded\"><h3 class=\"font-bold\">Hello from A</h3></article></div>\n\n)
		end
```

joeldrapper (changes requested):
I added a few points of feedback. On the whole, this looks great.

I wish we could find a better example for the Phlex component. Simply wrapping the other component in a Phlex component with the same name feels weird.

I wonder if it might be better to have Comment be a Phlex component which has a `ReactionComponent` (like an Emoji reaction to a comment) that's a ViewComponent component. And the reaction uses a Phlex component card. I’m struggling to come up with a better idea 🤷

nikolalsvk:
@joeldrapper I created the following tests:

1. ViewComponent renders a ViewComponent comment component that renders a Phlex reaction component
2. Phlex renders a Phlex comment component that renders a ViewComponent reaction component

I think in this case, we made examples that make sense + rendering of each other is properly tested.

WDYT?

As a side note, I got an error:

```
✗ Unhandled exception ActionView::Template::Error: no implicit conversion of Hash into String
```

When I tried to do `p @emoji, role: "img"`. Am I don't something wrong? 🤔

fixtures/dummy/app/views/comments/comment.rb
```diff
@@ -0,0 +1,25 @@
...
+
+module Views
+	module Comments
+		class Comment < Phlex::Component
+			def initialize(name:, body:)
+				@name = name
+				@body = body
+			end
+
+			def template(&block)
```
nikolalsvk, line 11: I tried using just `&` here, but the Rubocop is configured for Ruby 2.7 version and it throws errors.

We can have another issue/PR to support Ruby 3.1 version for Rubocop inside the project, WDYT?

joeldrapper: We use 2.7 style in the library and test files because we need to support Ruby 2.7. Documentation is built in 3.1.2 so we use the new style for that.

joeldrapper:
Thanks @nikolalsvk

PR #194, opened 2022-09-27
Rename `content` → `yield_content`

### LINKED ISSUE #190: Rename `content` method
I would like to allow people to use an expose the `content` method as a “slot” for their component. For example, you might want to define a component that has slots for title and content. For example, you could define a `Card` component with `title` and `content` slots:

```ruby
class Card < Phlex::Component
  def template(&)
    article(&)
  end

  def title(text)
    h3 text, class: "title"
  end

  def content(&block)
    div class: "content", &block
  end
end
```

And this component could be used like this:

```ruby
class Example < Phlex::Component
  def template
    render Card do |c|
      c.title "Hi"

      c.content do
        p "Hello!"
      end
    end
  end
end
```

Supporting this means renaming the `content` method defined here:

https://github.com/joeldrapper/phlex/blob/bb358e7942749d7834f679bb8c6e54115d5616cd/lib/phlex/component.rb#L58-L67

This method yields self to the block but also captures the output of that yield as text content if the buffer wasn't changed by the yield. It's difficult to know what to call it.

joeldrapper:
Closes #190

### DIFF
--- a/docs/components/tabs.rb
+++ b/docs/components/tabs.rb
@@ -8,7 +8,7 @@ module Components
 
 		def template(&block)
 			div class: "tabs flex flex-wrap relative my-5", role: "tablist" do
-				content(&block)
+				yield_content(&block)
 			end
 		end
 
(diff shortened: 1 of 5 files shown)


alexandreruban (approved):
LGTM!

PR #196, opened 2022-09-27
Fix templates page typo

henrik:


joeldrapper (approved):

joeldrapper:
👋 Thanks for the PR. I’ve just set things up so the docs are published from the latest release rather than the `main` branch so this will be going out with `0.3`.

PR #197, opened 2022-09-27
Create and document Phlex installation task for Rails

### LINKED ISSUE #184: Rails install generator
- [ ] Create a Phlex initialiser for configuration
- [ ] Configure Tailwind if installed
- [ ] Configure autoload paths to include `app` #197

alexandreruban:
Closes #184

In this PR, we create an installation task with for Rails. Currently, it only does one thing: Add `config.autoload_paths << Rails.root.join("app")` in `config/application.rb` to enable autoloading from `app/views` while preserving the `Views` namespace.

Please let me know if you want me to add an initializer / configuration for Tailwind.

### DIFF
--- a/lib/install/phlex.rb
+++ b/lib/install/phlex.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+say "Installing Phlex..."
+
+application_configuration_path = Rails.root.join("config/application.rb")
+application_configuration_content = File.read(application_configuration_path)
+
+pattern = %r(config.autoload_paths << (Rails.root.join\(.app.\)|.\#{root}/app.)\n)
+
+unless application_configuration_content.match?(pattern)
+	inject_into_class(
+		application_configuration_path,
+		"Application",
+		%(    config.autoload_paths << "\#{root}/app"\n)
+	)
+end
+
+say "Phlex successfully installed!"
--- a/lib/tasks/phlex_tasks.rake
+++ b/lib/tasks/phlex_tasks.rake
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+namespace :phlex do
+	desc "Install Phlex in the Rails application"
+	task :install do
+		install_file_path = File.expand_path("../install/phlex.rb", __dir__)
+
+		system "#{RbConfig.ruby} bin/rails app:template LOCATION=#{install_file_path}"
+	end
+end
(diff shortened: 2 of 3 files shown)


lib/install/phlex.rb
```diff
@@ -0,0 +1,18 @@
...
+application_configuration_path = Rails.root.join("config/application.rb")
+application_configuration_content = File.read(application_configuration_path)
+
+pattern = /config.autoload_paths << Rails.root.join\(["']app["']\)/
+
+unless application_configuration_content.match?(pattern)
+	inject_into_class(
+		application_configuration_path,
+		"Application",
+		%(    config.autoload_paths << Rails.root.join("app")\n)
```
joeldrapper, line 14: The Rails Guide [recommends](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#config-autoload-paths) using interpolation here. I don't know if that's supported in older versions of Rails though.

```suggestion
		%(    config.autoload_paths << "#{root}/app"
```

lib/tasks/phlex_tasks.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+namespace :phlex do
```
joeldrapper, line 3: Should this be a rake task or a generator? I think most Gems use generators, e.g. `bin/rails g phlex:install` but I might have that wrong as I always copy/paste. 🤔

joeldrapper (commented):
I added a couple of comments. We can always ship this and add the tailwind configuration in another PR.

lib/tasks/phlex_tasks.rake
alexandreruban: I think it depends: 

- [turbo-rails](https://github.com/hotwired/turbo-rails/blob/main/lib/tasks/turbo_tasks.rake) and [stimulus-rails](https://github.com/hotwired/stimulus-rails/blob/main/lib/tasks/stimulus_tasks.rake) use Rake tasks.
- [Devise](https://github.com/heartcombo/devise/blob/main/lib/generators/active_record/devise_generator.rb) use generators.

I think we can use one or the other. If you have a preference for generators, I can change for generators!

joeldrapper: Let's stick with tasks then.

joeldrapper (approved):

lib/install/phlex.rb
alexandreruban: I just tested and `config.autoload_paths << "#{root}/app"` works in a Rails 6.1 application!

I just made the change to replace `config.autoload_paths << Rails.root.join("app")\n)` with `config.autoload_paths << "#{root}/app"`

I won't merge this PR immediately as I'll add the Tailwind config before. I'll re-request your review when I'm ready 😃

PR #201, opened 2022-09-28
Add `helpers` proxy

mansakondo:
This PR adds the `helpers` proxy to be able to access helpers within a `Phlex::View`.

For example, you can use the `#t` helper provided in Rails views for translations:
```ruby
class Hello < Phlex::View
  delegate :t, to: :helpers
  
  def template
    h1 do
      t "hello"
    end
  end
end
```

### DIFF
--- a/docs/pages/rails_integration.rb
+++ b/docs/pages/rails_integration.rb
@@ -24,6 +24,30 @@ module Pages
 						end
 					end
 				RUBY
+
+				render Markdown.new(<<~MD)
+					## Helpers
+
+					You can use the `helpers` proxy to access helpers within a `Phlex::View`.
+
+					For example, you can use the `#t` helper for translations:
+				MD
+
+				render CodeBlock.new(<<~RUBY, syntax: :ruby)
+					# app/views/hello.rb
+
+					module Views
+						class Hello < Phlex::View
+						  delegate :t, to: :helpers
+
+							def template
+							  h1 do
+								  t "hello"
+								end
+							end
+						end
+					end
+				RUBY
 			end
 		end
 	end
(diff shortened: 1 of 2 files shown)


lib/phlex/view.rb
```diff
@@ -27,12 +27,15 @@ def compiled?
 			end
 		end
 
+    attr_reader :helpers
```
joeldrapper, line 30: Although an `attr_reader` is faster, I think we should probably just define this as a method and delegate to the `@_view_context` instance variable, and then inline that in a compilation step.

```suggestion
    def helpers
      @_view_context
    end
```

joeldrapper:
We should document this on the Rails Integration page but we can do that separately if you want to get this merged so you can continue with the `simple_calendar` port.

lib/phlex/view.rb
mansakondo: Do you want the method to be at the top of the public API?

mansakondo:
> We should document this on the Rails Integration page but we can do that separately if you want to get this merged so you can continue with the `simple_calendar` port.

No problem, we can do it now.

lib/phlex/view.rb
joeldrapper: I think it's fine here. It's tricky to know where to document this as it's not a Rails-specific feature, but it is a view_context specific feature so it works in Bridgetown, etc. I think for now, this is fine here but should be documented under Rails.

mansakondo:
How can I generate the Rails Integration page to see the result?

docs/pages/rails_integration.rb
```diff
@@ -24,6 +24,30 @@ def template
 						end
 					end
 				RUBY
+
+				render Markdown.new(<<~MD)
+				## Helpers
+
+				You can use the `helpers` proxy to access helpers within a `Phlex::View`.
+
+				For example, you can use the `#t` helper for translations:
```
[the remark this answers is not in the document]
mansakondo: Oops, my bad.

PR #202, opened 2022-09-28
Compiler improvements

### LINKED ISSUE #19: Static components
We should be able to compile logic-free components into methods returning a string with interpolation for the variables and contents.

joeldrapper:
Closes #19


lib/phlex/compiler/generators/standard_element.rb
```diff
@@ -7,20 +7,45 @@ class StandardElement
...
+					return unless positional_arguments
+
+					case positional_arguments.length
+					when 1
+						positional_arguments.first
+					when 0
+						nil
+					else
+						positional_arguments
+					end
```
mansakondo, line 43: I couldn't help but try the pattern-matching version, after watching the live stream :)
```ruby
case (arguments = @arguments&.parts)
in []
  nil
in [SyntaxTree::BareAssocHash, *]
  nil
in [first]
  first
else
  arguments
end
```

mansakondo, line 43: I couldn't help but try the pattern-matching version, after watching the live stream 😄 
```suggestion
					case (arguments = @arguments&.parts)
					in []
						nil
					in [SyntaxTree::BareAssocHash, *]
						nil
					in [first]
						first
					else
						arguments
					end
```

lib/phlex/compiler/visitors/statements.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Phlex::Compiler::Visitors
+	class Statements < Base
+		MUTATING_METHODS = [:raw, :whitespace, :comment, :yield_content, :text, :doctype]
```
joeldrapper, line 5: `yield_content` might not mutate.

PR #212, opened 2022-10-01
Isolate tests

joeldrapper:
What you gon' do when I appear?
W-W-when I premiere?
to isolate Rails when it shouldn't be there?

I was in the #212.


.github/workflows/ci.yml
```diff
@@ -20,8 +20,11 @@ jobs:
           ruby-version: ${{ matrix.ruby-version }}
           bundler-cache: true
 
-      - name: Tests
-        run: bundle exec sus
+      - name: Isolated tests
+        run: bundle exec sus test/isolated/**/*rb
+
+      - name: Rails tests
+        run: bundle exec sus test/rails/**/*rb
```
ParamagicDev, line 27: I think this glob is missing a file extension separator.

```suggestion
      - name: Isolated tests
        run: bundle exec sus test/isolated/**/*.rb

      - name: Rails tests
        run: bundle exec sus test/rails/**/*.rb
```

joeldrapper: I think the issue is `**/*` is interpreted differently in GitHub actions than it is on my local machine. Locally, it means “in this folder and/or sub folders”. In GitHub actions, it means “in sub folders”.

ParamagicDev: Perhaps this is a bash vs sh issue?

could try:

`/bin/bash -e "bundle exec sus test/isolated/**/*.rb"`

joeldrapper: That doesn't work either. 😞

PR #215, opened 2022-10-01
Allow tests to be ran against a single folder

bschrag620:
* adds list_tests executable for echoing a list of files
* updates to github actions yml to leverage list_tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,20 @@ jobs:
           ruby-version: ${{ matrix.ruby-version }}
           bundler-cache: true
 
-      - name: Tests
-        run: bundle exec sus
+      - name: Tests - compilation
+        run: bundle exec sus $(bin/list_tests phlex/compilation)
+
+      - name: Tests - compiler
+        run: bundle exec sus $(bin/list_tests phlex/compiler)
+
+      - name: Tests - integration
+        run: bundle exec sus $(bin/list_tests phlex/integration)
+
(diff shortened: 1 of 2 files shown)


joeldrapper:
It's only failing because of Rubocop. Looks like it's running all the tests in each case though. I wonder if we should just try to makes us take a folder, e.g. `bundle exec sus test/rails`. Looks like this is the executable. https://github.com/ioquatix/sus/blob/main/bin/sus

bschrag620:
>  Looks like it's running _all_ the tests in each case though. 

Typo in the command was attempting to run `list_test` instead of `list_tests`

> I wonder if we should just try to make sus take a folder, e.g. `bundle exec sus test/rails`. 

I had tried this locally but got errors that `test/<folder-name>` was a folder.

joeldrapper:
> I had tried this locally but got errors that test/<folder-name> was a folder.

Right, but I mean we could submit a PR to Sus to make this work.

bschrag620:
> Right, but I mean we could submit a PR to Sus to make this work.

Happy to look into that. Latest test suite does look like it properly ran each folder individually. Spot checking the total assertions real quick to make sure tests didn't get skipped.

bschrag620:
So I see a difference of 1 in total test files ran. There is a `rails.rb`  that is directly in `phlex/` which explains the difference.

joeldrapper:
> Happy to look into that.

In the meantime, I should merge this with #212.

PR #219, opened 2022-10-03
Form abstraction for Rails

joeldrapper:


lib/phlex/rails/form.rb
```diff
@@ -0,0 +1,49 @@
...
+
+module Phlex
+	module Rails
+		module Form
+			def initialize(model)
+				@model = model
+				@url = @view_context.url_for(@model)
+			end
+
+			def self.input_field(method_name, type:)
```
mansakondo, line 11: Is there a reason why this method isn't named or aliased to `field`?

mansakondo, line 11: Is there a reason why this method isn't named or aliased to `field`?
```suggestion
			def self.field(method_name, type:)
```

joeldrapper:
> I'm tempted to add some comments but I don't want to pollute your draft PR. Let me know if it's ok for me to comment on it.

Early comments can be really helpful. This was just a sketch when I posted it, I think it's mostly working now but still a long way to go.

mansakondo:
> > I'm tempted to add some comments but I don't want to pollute your draft PR. Let me know if it's ok for me to comment on it.
> 
> Early comments can be really helpful. This was just a sketch when I posted it, I think it's mostly working now but still a long way to go.

Yeah, since it was a sketch when I tried to test the PR I had to make some changes to make it work. That's why I wanted to add some comments, but then I realized that it was unnecessary in this case.

lib/phlex/rails/form.rb
```diff
@@ -0,0 +1,56 @@
...
+				input name: "commit", type: "submit", value: value
+			end
+
+			def url
+				@_view_context.url_for(@model)
+			end
+
+			def field_name(field)
+				@_view_context.field_name(ActiveModel::Naming.param_key(@model.class), field)
+			end
```
dhnaranjo, line 39:
```suggestion
			def field_name(*field)
				@_view_context.field_name(ActiveModel::Naming.param_key(@model.class), *field)
			end
```

Allows `field_name(:its, :ya, index, :boy) => form[its][ya][0][boy]`

joeldrapper:
Shipping this as undocumented experimental.

PR #222, opened 2022-10-05
Support html safety for the #render_in method

bradgessler:
Fixes issue where this code:

```ruby
  def show
    render inline: (Views::Layouts::Article.new.render_in(view_context){ "<p>Hi</p>".html_safe }), content_type: "text/html"
  end
```

Would result in a rendition like this:

193955078-810f8309-c9e2-4261-a97d-683b28d2f580

The updated method checks if the string is `html_safe` before deciding to call `raw` or `text`


lib/phlex/renderable.rb
```diff
@@ -26,7 +26,9 @@ def render_in(view_context, &block)
 						output = yield(*args, **kwargs)
 						unchanged = (original_length == @_target.length)
 
-						text(output) if unchanged && output.is_a?(String)
+						if unchanged && output.is_a?(String)
+							output.html_safe? ? raw(output) : text(output)
+						end
```
joeldrapper, line 31: Calling `html_safe` will be unacceptably slow, but I’m happy to use a case statement to check for the `ActiveSupport::SafeBuffer` type. We just need to figure out how to handle `active_support` not being around, since it's not a dependency of Phlex.

```suggestion
						if unchanged
						  case output
						  when ActiveSupport::SafeBuffer
						  	raw(output)
						  when String
						  	text(output)
						  end
						end
```

joeldrapper (changes requested):

bradgessler:
Just tried your proposed changes in a Rails project and it works. 

The "dumbest" thing I can think of to make this work outside of `ActiveSupport` is a `#raw_render_in` method. If that's gross then maybe I'd propose an `Phlex.raw` method that wraps a string in a `Phlex::RawString` (I don't think I got that name right) class that could then be checked by Phlex.

Putting that all together, the call would look something like this outside of Rails:

```ruby
Views::Layouts::Article.new.render_in { Phlex.raw("<p>Hi</p>") }
```

joeldrapper:
I’d be happy to have some kind of ActiveSupport feature detection that includes a module to override this method. 🤔

lib/phlex/renderable.rb
```diff
@@ -26,7 +26,14 @@ def render_in(view_context, &block)
...
 						unchanged = (original_length == @_target.length)
 
-						text(output) if unchanged && output.is_a?(String)
+						if unchanged
+							case output
+							when ActiveSupport::SafeBuffer
+								raw(output)
+							when String
+								text(output)
+							end
```
joeldrapper, line 35:
```suggestion
							if defined?(ActiveSupport::SafeBuffer) && output.is_a?(ActiveSupport::SafeBuffer)
								raw(output)
							else
								text(output)
							end
```

joeldrapper (approved):

joeldrapper:
👋 I thought of a way to add feature detection and it seems to work.

PR #224, opened 2022-10-05
Initial i18n support

joeldrapper:
Add support for I18n translations through the `translate` and `t` methods on views that include `Phlex::I18n`.

### DIFF
--- a/lib/phlex/i18n.rb
+++ b/lib/phlex/i18n.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Phlex
+	module I18n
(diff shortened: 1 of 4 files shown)


lib/phlex/i18n.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Phlex
+	module I18n
```
joeldrapper, line 4: I’m not sure if I like `I18n` as a module name here. Perhaps `Translations` would be better.

joeldrapper:
Closing this in favour of #251.

PR #233, opened 2022-10-06
Add `<hr>` void element tag

jai-x:


joeldrapper (approved):

joeldrapper:
Thanks for adding this. In the meantime, you can just add `register_element :hr` in your ApplicationView and have your other views inherit from that.

PR #246, opened 2022-10-11
Conditional token array bug fixed

mbriggs:
Just a small bug fix, love the project, love the direction you are taking it :-)

I kind of feel like it would be easy to support the symbol / string / array stuff on conditional tokens globally. If you agree I could do a quick followup pr

### DIFF
--- a/lib/phlex/view.rb
+++ b/lib/phlex/view.rb
@@ -165,7 +165,7 @@ module Phlex
 				case token
 				when Symbol then tokens << token.name
 				when String then tokens << token
-				when Array then tokens.concat(t)
+				when Array then tokens.concat(token)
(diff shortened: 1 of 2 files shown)


joeldrapper:
Looks great. Rubocop is wrong, I’m going to remove that cop.

joeldrapper:
> I kind of feel like it would be easy to support the symbol / string / array stuff on conditional tokens globally. If you agree I could do a quick followup pr

What do you mean by supporting these globally?

mbriggs:
oh never mind! I didn't realize that `join` was recursive, just assumed the logic about adding conditional tokens was not happening with non conditional tokens

PR #251, opened 2022-10-12
Add I18n support

### LINKED ISSUE #179: Better i18n support
Currently Phlex does not provide anything for i18n, one can include `I18n` and then use the translation helpers with absolute keys.

So while this works, adding the rails helpers for i18n will cause problems, when trying to use relative keys like `t(".new")`, it will raise an error, that there is no `@virtual_path` set.

ViewComponent solves this, by setting the `@virtual_path` in the `render_in` method. 

ViewComponent also supports having the language files as sidecar files, which is pretty neat, and might be a cool thing for Phlex as well.

Some of my research on the topic can be found here: https://discord.com/channels/629472241427415060/1002702966551216139/1022976024432095373

stephannv:
Following up the PR https://github.com/joeldrapper/phlex/pull/224, I added support to relative keys (starting with period). I got inspiration from ViewComponent (https://github.com/ViewComponent/view_component/blob/main/lib/view_component/base.rb#L505-L510).
I added tests to ensure the virtual path is being filled with correct value, but I had to use component from dummy application.
I added tests to ensure the translation is working with relative keys and non-relative keys.
I don't know if the strategy I used is the best, but if you want changes, you can request, or if you want to discard this PR, that's fine too.
Great component framework! Thanks for the effort.

Closes #179


lib/phlex/i18n.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Phlex
+	module I18n
```
joeldrapper, line 4: Can we rename this `Translation`?

joeldrapper (changes requested):
Thanks so much for working on this. This is going to come in really handy!

I wonder if you might be able to add a small section to the docs under the `Helpers` page? But also happy to do that in a follow-up another time.

I've a few minor points of feedback but love the general direction. 👍

joeldrapper:
@RolandStuder what are your thoughts on this?

stephannv:
> Thanks so much for working on this. This is going to come in really handy!
> 
> I wonder if you might be able to add a small section to the docs under the `Helpers` page? But also happy to do that in a follow-up another time.
> 
> I've a few minor points of feedback but love the general direction. 👍

This weekend I will make the necessary adjustments. Thanks for the feedback.

fixtures/dummy/app/views/application_view.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module Views
```
stephannv, line 3: I created this because I had to add Translation to Articles::Form view, so I took the opportunity to create something more similar to recommended usage from documentation/generator.
But I can remove it if this isn't wanted.

lib/phlex/configuration.rb
```diff
@@ -2,6 +2,10 @@
 
 module Phlex
 	class Configuration
-		# Config coming soon.
+		attr_accessor :phlex_path
```
stephannv, line 5: I was in doubt about that name or `views_path`. If you have some suggestion I will accept it.

lib/phlex/translation.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Phlex
+	module Translation
+		def self.included(view)
+			view.extend(ClassMethods)
+		end
+
+		module ClassMethods
+			attr_writer :translation_path
```
stephannv, line 10: I moved the `virtual_path` logic to translation, and renamed it to `translation_path`, but if you think this information will be useful to other features, I can extract it to another module and rename back to `virtual_path`.
I decided to go with the memoization instead `included`/`inherited`, because the translation_path evaluation code will run only if `translate` or `t` methods are called, this can help the framework not degrade the performance.

lib/phlex/translation.rb
```diff
@@ -0,0 +1,34 @@
...
+
+			def translation_path
+				@translation_path ||= begin
+					source_location = const_source_location(name).first # Get view absolute file path
+					relative_path = source_location.split(Phlex.configuration.phlex_path).last # Get relative path to the phlex path
+					relative_path.delete_prefix("/").delete_suffix(".rb").tr("/", ".")
```
stephannv, line 16: The suggestion was to use `#[]`, but i ended up doing it another way. If you have a suggestion on how to improve this using `#[]` i will accept it.

lib/phlex/translation.rb
```diff
@@ -0,0 +1,34 @@
...
+				rescue StandardError
+					# When class doesn't have file location, eg. Class.new(Phlex::View), the const_source_location doesn't work,
+					# it raises exception. This rescue ensures the `translate`  will work in this situation.
+					# `t("hello")` or `t(".hello")` will lookup for "en.hello" translation
+					""
```
stephannv, line 21: In the future, we can add a warning here:
```suggestion
          Phlex.configuration.logger.warn "Cannot evaluate view translate path ...`
					""
```

docs/pages/translations.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+module Pages
+	class Translations < ApplicationPage
```
stephannv, line 4: I added a new page instead an small section, because Translation may end up having more features (eg. Generating `.yml` files).

English is not my primary language, so there must be errors in the text, and there must also be better ways of writing, so can you help me and suggest writings, I will be grateful.

lib/phlex/translation.rb
joeldrapper, line 16: @RolandStuder posted this related comment on the commit:

> You are relying on file paths here, though people might have custom autoloading setups, and have components / views outside of the `/app/views` directory. I think the better way would be to use `class.name.underscore` and use that as the "path", this would be more consistent with what i18n generally does, and it would be more flexible.
> 
> What do you think?

joeldrapper: I agree, `class.name.underscore` would be better. We could update the configuration to take a namespace to be removed from the beginning of the `class.name`.

lib/phlex/translation.rb
```diff
@@ -0,0 +1,25 @@
...
+	module Translation
+		def self.included(view)
+			view.extend(ClassMethods)
+		end
+
+		module ClassMethods
+			attr_writer :translation_path
+
+			def translation_path
+				@translation_path ||= name&.split("::")&.join(".")&.downcase.to_s
```
stephannv, line 13: This will use full class name, for example:
`Views::Layouts::Navbar` => `views.layouts.navbar`. I think it's fine to use `views.` prefix (later we can add a setting to ignore it), but the user has the option to do this at application.rb `config.autoload_paths << "#{root}/app/views"` instead `config.autoload_paths << "#{root}/app"`, and define Phlex views without `Views` module like it did in my project: 
- https://github.com/stephannv/nindika/blob/main/config/application.rb#L25
- https://github.com/stephannv/nindika/blob/main/app/views/layouts/navbar_component.rb

Ps.: 
- `sub` and `str["::"] = "."` didn't work here because it replaces just the first occurrence.
- `split("::").join(".")` is faster than `gsub("::", ".")` according to benchmark I did in my machine.
Screen Shot 2022-10-22 at 12 21 50

lib/phlex/configuration.rb
```diff
@@ -2,6 +2,10 @@
...
+
+		def initialize
+			@phlex_path = "app/views"
+		end
```
joeldrapper, line 9: Can we remove this now?

stephannv: Yes. I will remove it.

joeldrapper (approved):

PR #252, opened 2022-10-13
Experimental view methods

joeldrapper:
Experimenting with an alternative to `render ComponentName.new`, where `ComponentName` is a method.

Before:
```ruby
render Modal::Wrapper::new do
  render Modal::Body::new do
    p "Do you really want to continue?"
  end

  render Modal::Actions::new do
    render(Buttons::CancelButton.new) { "Oh no!" }
    render(Buttons::ApproveButton.new) { "Absolutely!" }
  end
end
```

After:
```ruby
Modal_Wrapper do
  Body do
    p "Do you really want to continue?"
  end

  Actions do
    Buttons_CancelButton { "Oh no!" }
    Buttons_ApproveButton { "Absolutely!" }
  end
end
```

### DIFF
--- a/lib/phlex/renderable.rb
+++ b/lib/phlex/renderable.rb
@@ -8,7 +8,10 @@ module Phlex
 					block = Phlex::Block.new(self, &block)
 				end
 
+				@_original_wrapper = @_wrapper
+				@_wrapper = renderable
 				renderable.call(@_target, view_context: @_view_context, parent: self, &block)
+				@_wrapper = @_original_wrapper
 			elsif renderable.is_a?(Class) && renderable < View
 				raise ArgumentError, "You tried to render the Phlex view class: #{renderable.name} but you probably meant to render an instance of that class instead."
 			else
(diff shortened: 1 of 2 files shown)


joeldrapper:
The one thing I’m really not sure about is the wrapper-relative constant lookup. We never scope anything else to the wrapper. One alternative is to use the yielded wrapper view to call relative children, e.g.

```ruby
Modal_Wrapper do |m|
  m.Body do
    p "Do you really want to continue?"
  end

  m.Actions do
    Buttons_CancelButton { "Oh no!" }
    Buttons_ApproveButton { "Absolutely!" }
  end
end
```

dhnaranjo:
I dig this most of all because I am annoyed by needing to use parentheses on block-receiving `render` calls.

joeldrapper:
> I dig this most of all because I am annoyed by needing to use parentheses on block-receiving `render` calls.

Same! 😆

Issue #258, opened 2022-10-14
Avoid template naming conflict


joelmoss:
Have you thought about renaming the `template` method to avoid conflicting with the HTML `<template>` tag? For example, `render`?

This would completely remove that gotcha/caveat, and also makes a little more sense from a naming POV.

joeldrapper:
I have thought about it a lot and I haven’t been able to come up with a good name that isn't taken. The instance method `render` is used to render other things inside the template. `call` is part of the external API so doesn't make sense to define that directly. I’m not a fan of `build` or `perform`.

evenreven:
Maybe it's because I know Vue a little, but because of Vue there's kind of a precedent for using template as the markup generating block in single file components. Sure, it's more correct in javascript land, but it still feels conceptually right to me. (Also, I think the conflict with the template element is pretty niche unless you use the template tag a lot, and when you write HTML with a Ruby DSL, I think adjustments are to be expected anyway.)

joeldrapper:
It's worth noting that it is possible to correct `template` calls from inside the template method using the compiler. I just don't know if I want to go down that path because then it's not really Ruby anymore. It's a new language that operates slightly outside of the Ruby rules.

joeldrapper:
I think this might be a way out. https://github.com/joeldrapper/phlex/pull/260

joelmoss:
Oh nice! 👏

And as a happy side note, I think this should also work for #259...

```ruby
# ...
if block.instance_variable_get(:@rendering)
  respond_to?(:before_render) && before_render
  super(&block)
else
# ...
```

I still can't help but thinking that `render` would be a better name for the `template` method, and rename the internal methods that conflict.  But the above is ✨

joeldrapper:
`render` is a verb, it's a "so something" method. `template` _conceptually_ doesn't do anything, it is something.

As things stand, we go through three methods to render a template and I think they're well named: when I _render_ a view, the view is _called_ and outputs its _template_.

joelmoss:
When you describe it like that, it makes sense, but as far as a public API is concerned, the `template` method is all that matters, and I see that as I am rendering the content of the method that is currently called `template`. So a verb makes sense.

But hey, each to their own. You've already solved the template HTML tag conflict, so I'm happy 👏

thx

joeldrapper:
Naming is hard.

joelmoss:
Oh don't remind me 😫

joeldrapper:
[...]

PR #260, opened 2022-10-14
Magic template method

joeldrapper:
To avoid the template naming conflict (#258) we want `template` to behave differently if it is called by `call` to render the view or if it's called from inside the `template` method to insert a `<template>` tag.

By tainting the block that `call` passes to `template`, we have the opportunity for a prepended `template` override to check whether it was called by `call` and passed this tainted block. The override can either call `super` or insert a `<template>` tag.

Something similar was attempted before using instance variables, but it caused an issue: when subclassing another view, prepends would get stacked up in the inheritance tree and `super` would just point to the previous override.

### DIFF
--- a/lib/phlex/view.rb
+++ b/lib/phlex/view.rb
@@ -9,6 +9,16 @@ module Phlex
 		extend HTML
 		include Renderable
 
+		module TemplateOverride
+			def template(**kwargs, &block)
+				if block.instance_variable_get(:@rendering)
+					super(&block)
+				else
+					template_tag(**kwargs, &block)
+				end
+			end
+		end
+
 		class << self
 			attr_accessor :rendered_at_least_once
 
@@ -25,6 +35,10 @@ module Phlex
 			def compiled?
 				!!@compiled
 			end
+
+			def inherited(child)
+				child.prepend(TemplateOverride)
+			end
 		end
 
 		def call(buffer = +"", view_context: nil, parent: nil, &block)
@@ -36,6 +50,8 @@ module Phlex
 			@_parent = parent
 			@output_buffer = self
 
+			block ||= -> {}
+			block.instance_variable_set(:@rendering, true)
 			template(&block)
 
 			self.class.rendered_at_least_once ||= true


joeldrapper:
This seems to work pretty well and it's okay performance-wise. The next step is to figure out how to compile it. Ideally, when compilation is enabled, `TemplateOverride` will never be prepended.

lib/phlex/view.rb
```diff
@@ -9,6 +9,16 @@ class View
 		extend HTML
 		include Renderable
 
+		module TemplateOverride
+			def template(**kwargs, &block)
+				if block.instance_variable_get(:@rendering)
```
joeldrapper, line 14: If the block has the `@rendering` instance variable, we know this was called from `#call` and the intention is to render the view.

lib/phlex/view.rb
```diff
@@ -9,6 +9,16 @@ class View
...
+					super(&block)
```
joeldrapper, line 15: If the block has the `@rendering` instance variable, we know this was called from `#call` and the intention is to render the view, so we pass the block to `super`.

lib/phlex/view.rb
```diff
@@ -25,6 +35,10 @@ def compile
 			def compiled?
 				!!@compiled
 			end
+
+			def inherited(child)
+				child.prepend(TemplateOverride)
+			end
```
joeldrapper, line 41: When `Phlex::View` is inherited, we want to prepend the `TemplateOverride` module on the child so that `template` is defined by our override.

lib/phlex/view.rb
```diff
@@ -36,6 +50,8 @@ def call(buffer = +"", view_context: nil, parent: nil, &block)
 			@_parent = parent
 			@output_buffer = self
 
+			block ||= -> {}
+			block.instance_variable_set(:@rendering, true)
```
joeldrapper, line 54: Ensure we always have a block, even if it doesn't do anything and set the instance variable `@rendering` on it.

joeldrapper:
I’m closing this. There’s some additional info here detailing what was tried, what worked and what didn't. https://www.theabstraction.space/p/magic-template-methods

I think the renaming approach could work on the class rather than singleton classes if everything was eager loaded already before we did the rename. However, I expect that would interfere with Zeitwerk reloading.

PR #263, opened 2022-10-15
Document the ability to register custom elements

mansakondo:
This PR adds documentation to allow users to register custom elements.

### DIFF
--- a/docs/pages/views.rb
+++ b/docs/pages/views.rb
@@ -140,7 +140,7 @@ module Pages
 				render Example.new do |e|
 					e.tab "hello.rb", <<~RUBY
             class Hello < Phlex::View
-              register_element :hello
+              register_element :trix_editor
 
               def template
                 hello do
--- a/docs/pages/views.rb
+++ b/docs/pages/views.rb
@@ -130,6 +130,34 @@ module Pages
 					end
 					```
 				MD
+
+				render Markdown.new(<<~MD)
+					## Registering custom elements
+
+					Here's how you can register custom elements.
+				MD
+
+				render Example.new do |e|
+					e.tab "hello.rb", <<~RUBY
+            class Hello < Phlex::View
+              CUSTOM_ELEMENTS = {
+                hello: "hello"
+              }.freeze
+
+              CUSTOM_ELEMENTS.each do |method_name, tag|
+                register_element(method_name, tag)
+              end
+
+              def template
+                hello do
+                  "World!"
+                end
+              end
+            end
+					RUBY
+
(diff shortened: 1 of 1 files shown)


docs/pages/views.rb
```diff
@@ -130,6 +130,34 @@ def template(&content)
...
+				render Example.new do |e|
+					e.tab "hello.rb", <<~RUBY
+            class Hello < Phlex::View
+              CUSTOM_ELEMENTS = {
+                hello: "hello"
+              }.freeze
+
+              CUSTOM_ELEMENTS.each do |method_name, tag|
+                register_element(method_name, tag)
+              end
```
joeldrapper, line 149: I think the golden path should be to just do this:

```suggestion
              register_element :element_name
```

docs/pages/views.rb
```diff
@@ -130,6 +130,34 @@ def template(&content)
 					end
 					```
 				MD
+
+				render Markdown.new(<<~MD)
+					## Registering custom elements
+
+					Here's how you can register custom elements.
```
joeldrapper, line 137:
```suggestion
					Here's how you can register custom elements. The custom element will only be available in the view where it is registered and subclasses of that view.
```

docs/pages/views.rb
mansakondo: Indeed, that's way cleaner.

docs/pages/views.rb
```diff
@@ -130,6 +130,28 @@ def template(&content)
...
+				render Markdown.new(<<~MD)
+					## Registering custom elements
+
+					You can register custom elements with the `register_element` macro. The custom element will only be available in the view where it is registered and subclasses of that view.
+				MD
+
+				render Example.new do |e|
+					e.tab "hello.rb", <<~RUBY
+            class Hello < Phlex::View
+              register_element :hello
```
joeldrapper, line 143: HTML custom elements must contain a dash, so let's include that in our example.

```suggestion
              register_element :trix_editor
```

docs/pages/views.rb
```diff
@@ -130,6 +130,28 @@ def template(&content)
...
+
+              def template
+                hello do
+                  "World!"
+                end
```
joeldrapper, line 148:
```suggestion
                trix_editor input: "content" autofocus: true
```

mansakondo:
Ah, sorry I should have fixed the indentation myself

joeldrapper:
No problem. Rubocop is really bad at fixing indentation because we use tabs instead of spaces.

Issue #264, opened 2022-10-15
RFC: Attribute methods


joelmoss:
First of all, really starting to enjoy Phlex - I think you may be onto the something here 👍

This is an idea that was borne out of my desire (need?) to automatically modify certain attributes of any HTML tag. Primarily, I want to be able to "rewrite" class names to support CSS modules. But this could be used for any attribute with simple and quite advanced uses.

Define an `#[name]_attribute` method in your view, and then any attribute in that view will be processed through this method.

So as a first trivial example, I want to be able to pass an array of class names instead of a string (yes I know I can use the `classes` or `token` helpers, but I want to save keystrokes - I don't have many left 😜).

```ruby
class HeaderView < Phlex::View
  # Returns <div class="some class names"></div>
  def template
    div class: ['some', 'class', 'names']
  end

  def class_attribute(value)
    value.join ' '
  end
end
```

The `class_attribute` method receives the value given in `template` (`['some', 'class', 'names']`), and returns the result of joining the array. But of course, it could do absolutely anything.

Any method that begins with an attribute name and ends with `_attribute`, can be used to "massage" that attributes value. not convinced on the naming here, but it's a start.

There is also some scope here to specify which tags an attribute method should be applied to, perhaps like this:

```ruby
  def class_attribute(value, tag)
    value.join ' ' if tag == :div
  end
```

Anyway, this is just an idea, but also a simple one that would need a very small change to `Phlex::View#_build_atttibutes`:

```ruby
# ...
  if respond_to?(:"#{k}_attribute")
    v = send :"#{k}_attribute", v
  end
# ...
```

Thoughts?

joelmoss:
After a little testing, I just noticed that my change to `_build_attributes` won't work as is, because attributes are cached :( So testing other implementations.

joeldrapper:
Hey Joel, thanks for writing this up. I can see how this could be useful, but also I have thoughts:

First, we're only saving 2 keystrokes in your example. But you can save 9 keystrokes by using a String instead:

```ruby
div class: ['some', 'class', 'names'] # 37
div **classes('some', 'class', 'names') # 39
div class 'some class names' # 28
```

You also save memory allocations and operations:

```ruby
div class: ['some', 'class', 'names'] # 1 Array, 3 + 1 Strings and a join
div **classes('some', 'class', 'names') # 1 Hash, 1 Array, 3 + 1 Strings and a join
div class 'some class names' # 1 String and no join
```

[...]

PR #265, opened 2022-10-15
feat: #264 Support "attribute methods"

joelmoss:
Not been able to figure out how to test the cache busting :(

### DIFF
--- a/lib/phlex/view.rb
+++ b/lib/phlex/view.rb
@@ -200,8 +200,6 @@ module Phlex
 		end
 
 		def _build_attributes(attributes, buffer:)
-			attr_method_used = false
-
 			attributes.each do |k, v|
 				next unless v
 
@@ -232,7 +230,7 @@ module Phlex
 				end
 			end
 
-			attr_method_used
+			buffer
 		end
 	end
 end
--- a/lib/phlex/view.rb
+++ b/lib/phlex/view.rb
@@ -183,9 +183,10 @@ module Phlex
 				attributes[:href] = attributes[:href].sub(/^\s*(javascript:)+/, "")
 			end
 
-			_build_attributes(attributes, buffer: buffer)
+			attr_method_used = _build_attributes(attributes, buffer: buffer)
 
-			unless self.class.rendered_at_least_once
+			# Don't cache if an attribute method was used.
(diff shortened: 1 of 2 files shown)


joeldrapper:
We can probably just do the transform before generating the cache keys. I'll take a look at this next week.

lib/phlex/view.rb
```diff
@@ -178,13 +178,20 @@ def helpers
...
+		def resolve_attributes(attributes)
+			attributes
+		end
+
 		def _attributes(attributes, buffer: +"")
 			if attributes[:href]&.start_with?(/\s*javascript/)
 				attributes[:href] = attributes[:href].sub(/^\s*(javascript:)+/, "")
 			end
 
+			attributes = resolve_attributes(attributes)
```
joeldrapper, line 190: I think `tailor_attributes` might be a better name for this, “to make or adapt for a particular purpose”.

Unfortunately, we have performance to contend with once again. A method call or `responds_to?` check will add about 1% overhead, which is quite a lot if you never use the feature.

Perhaps this could go in a module, like the callbacks.

joelmoss: There is no `responds_to?` call here.

joeldrapper: I know, but that's the alternative to a method call. They're both about 1% I believe.

joelmoss: Oh right, I get ya. 

Yeah, a module could be better actually, as it makes it completely opt in, with no perf hit if not used.

joeldrapper: The only problem is we can't make a module that overrides this without having to duplicate the rest of the method in that module.

joelmoss: Then I see the only option being to always call `resolve_attributes`

joeldrapper: There is another option that doesn't cost anything if you don't use it. We could rename `_attributes` to `resolve_attributes` or something like that and let you override it. You’d have to pass your result to `super` but I think that's reasonable for something like this.

joelmoss: Ok, so `resolve_attributes`, `attributes`, `tailor_attributes`? Maybe just `attributes`?

PR #270, opened 2022-10-17
Another crack at "Experimental view methods"

dhnaranjo:
Whattup @joeldrapper I liked https://github.com/joeldrapper/phlex/pull/252 and I think we should keep trying.

This strategy depends on eager loading to build a lookup where a component registers itself at every level of their namespace on up, and callers of components in this style search for the component they're looking for from their namespace on up. We'll need to use the naive lookup in your original PR when eager loading is disabled.

We pay the cost up front at app start because who cares? Plus it can be optional behavior. All you gotta do is drop the module into your component and you're good.

I'm not sure what kind of benchmarks your efforts didn't pass but why not give them a try here?

### DIFF
--- a/lib/phlex/renderable.rb
+++ b/lib/phlex/renderable.rb
@@ -8,7 +8,10 @@ module Phlex
 					block = Phlex::Block.new(self, &block)
 				end
 
+				@_original_wrapper = @_wrapper
+				@_wrapper = renderable
 				renderable.call(@_target, view_context: @_view_context, parent: self, &block)
+				@_wrapper = @_original_wrapper
 			elsif renderable.is_a?(Class) && renderable < View
 				raise ArgumentError, "You tried to render the Phlex view class: #{renderable.name} but you probably meant to render an instance of that class instead."
 			else
--- a/lib/phlex/view.rb
+++ b/lib/phlex/view.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require "benchmark"
 
 if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("3.0")
 	using Overrides::Symbol::Name
@@ -42,6 +43,15 @@ module Phlex
 
 			buffer
 		end
+
+		def method_missing(name, *args, **kwargs, &block)
+			view_class = ComponentRegistry.instance.fetch(name, self.class)
+
+			# we can include the slow method as fallback / when eager loading is disabled, omitted while trying it out
+
+			render view_class.new(*args, **kwargs), &block
+		end
+
 
 		def rendered?
 			@_rendered ||= false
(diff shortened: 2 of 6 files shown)


lib/phlex/view.rb
```diff
@@ -43,6 +44,16 @@ def call(buffer = +"", view_context: nil, parent: nil, &block)
 			buffer
 		end
 
+		def method_missing(name, *args, **kwargs, &block)
+			view_class = ComponentRegistry.instance.fetch(name, self.class)
+
+			# we can include the slow method as fallback / when eager loading is disabled, omitted while trying it out
+			view_class ||= (@_wrapper || self).instance_eval name.to_s.gsub("_", "::")
```
joeldrapper, line 51: This never actually worked for non eager-loaded views. 😞

lib/phlex/view.rb
```diff
@@ -43,6 +44,16 @@ def call(buffer = +"", view_context: nil, parent: nil, &block)
 			buffer
 		end
 
+		def method_missing(name, *args, **kwargs, &block)
```
joeldrapper, line 47: `method_missing` and the `*args` and `**kwargs` splats are the slow bit. I wanted to use the `inherited` hook to eagerly define methods so `method_missing` wouldn't need to be hit.

lib/phlex/renderable.rb
```diff
@@ -8,7 +8,10 @@ def render(renderable, *args, **kwargs, &block)
 					block = Phlex::Block.new(self, &block)
 				end
 
+				@_original_wrapper = @_wrapper
+				@_wrapper = renderable
 				renderable.call(@_target, view_context: @_view_context, parent: self, &block)
+				@_wrapper = @_original_wrapper
```
joeldrapper, line 14: I’m not convinced it should be `@wrapper` relative if we do this. https://github.com/joeldrapper/phlex/pull/252#issuecomment-1277362082

joeldrapper (commented):
This is really cool. I need to take some time to look through your implementation. I’ve been thinking I should revisit this ever since I closed the last PR. There's definitely something in it and I think the compiler will be able to remove any performance issues.

lib/phlex/view.rb
joeldrapper: I had a conversation about this with Xavier Noria yesterday https://twitter.com/joeldrapper/status/1581765885165178880

dhnaranjo:
Alright I really went in on this thing and came up with something pretty nice. I liked your idea in #252 to yield the parent's context, but I wanted to not need to define methods on every single potential caller. Honestly, I kinda beat the problem into submission towards the end and its maybe the bad kinda clever, I'd be jazzed if someone could clean it up some.

https://gist.github.com/dhnaranjo/3decca0782374aaa7f2200d69bd710ee

joeldrapper:
I think we'll find a good alternative to this in a declarative web components abstraction. Since each component will have a corresponding custom HTML element, you’ll be able to use that element in place of rendering a view.

PR #275, opened 2022-10-20
WIP: Add `render_view` test helper

stephannv:
A first attempt to add test helpers. What do you think? 
The documentation is still pretty sketchy, I didn't try the helper with Minitest yet, and it is too attached with Rails.

I'm extracting this from my project:
The helper method:
https://github.com/stephannv/nindika/blob/main/spec/support/phlex.rb#L3-L9

Usage:
https://github.com/stephannv/nindika/blob/main/spec/views/layouts/navbar_component_spec.rb#L6-L10

With this helper (and capybara) will be possible to do something like:
```ruby
result = render_view(my_view.new) 
expect(result).to have_link("Home", href: "/")
```

### DIFF
--- a/docs/components/layout.rb
+++ b/docs/components/layout.rb
@@ -51,6 +51,7 @@ module Components
 										render Nav::Item.new("Views", to: Pages::Views, active_page: @_parent)
 										render Nav::Item.new("Templates", to: Pages::Templates, active_page: @_parent)
 										render Nav::Item.new("Helpers", to: Pages::Helpers, active_page: @_parent)
+										render Nav::Item.new("Testing", to: Pages::Testing, active_page: @_parent)
 									end
 
 									h2(class: "text-lg font-semibold pt-5") { "Rails" }
--- a/lib/phlex/test_helpers.rb
+++ b/lib/phlex/test_helpers.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Phlex
+	module TestHelpers
+		def render_view(view, &block)
+			rendered = controller.view_context.render(view, &block)
+			Nokogiri::HTML.fragment(rendered)
+		end
+
+		def controller
+			ActionView::TestCase::TestController.new
+		end
+	end
+end
--- a/docs/pages/testing.rb
+++ b/docs/pages/testing.rb
@@ -1,0 +1,94 @@
+# frozen_string_literal: true
+
+module Pages
+	class Testing < ApplicationPage
+		def template
+			render Layout.new(title: "Testing") do
+				render Markdown.new(<<~MD)
+					# Testing
+
+					## RSpec
+
+					Add test helpers to your suite:
+
+					```ruby
+					# spec/support/phlex.rb
+					require "phlex/test_helpers"
+
+					RSpec.configure do |config|
+						config.include Phlex::TestHelpers, type: :view
+					end
+					```
+
+					Remember to require this file in `rails_helper.rb` uncommenting the following line:
+
+					```ruby
+					Dir[Rails.root.join("spec/support/**/*.rb")].each { |f| require f }
+					```
+
+					### Asserting without capybara
+
+					```ruby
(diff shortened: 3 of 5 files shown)


lib/phlex/test_helpers.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module Phlex
+	module TestHelpers
+		def render_view(view, &block)
+			rendered = controller.view_context.render(view, &block)
+			Nokogiri::HTML.fragment(rendered)
+		end
+
+		def controller
+			ActionView::TestCase::TestController.new
```
joeldrapper, line 11: Should we memoize this?

```suggestion
			@controller ||= ActionView::TestCase::TestController.new
```

lib/phlex/test_helpers.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Phlex
+	module TestHelpers
+		def render_view(view, &block)
```
joeldrapper, line 5: I wonder if we should use `render` here instead. What do you think? I’m not sure if that’s defined by anything else.

```suggestion
		def render(view, &block)
```

lib/phlex/test_helpers.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Phlex
+	module TestHelpers
```
joeldrapper, line 4: We should either move this to `Phlex::Rails::TestHelpers` or make these not depend directly on Rails. For example, we could use `call` directly if `ActionView::TestCase::TestController` isn't defined.

joeldrapper (commented):
Love this direction!

I accidentally marked the PR as ready for review and can't convert it back to a draft agin. Sorry about that!

docs/components/layout.rb
```diff
@@ -51,6 +51,7 @@ def template(&block)
 										render Nav::Item.new("Views", to: Pages::Views, active_page: @_parent)
 										render Nav::Item.new("Templates", to: Pages::Templates, active_page: @_parent)
 										render Nav::Item.new("Helpers", to: Pages::Helpers, active_page: @_parent)
+										render Nav::Item.new("Testing", to: Pages::Testing, active_page: @_parent)
```
joeldrapper, line 54: Shall we add a "Testing" section with "RSpec" and "Minitest" pages?

lib/phlex/test_helpers.rb
stephannv: RSpec defines `render` method for view tests: https://github.com/rspec/rspec-rails/blob/main/lib/rspec/rails/example/view_example_group.rb#L66

joeldrapper: Hmm, I think we might even be able to use their definitions of render.

stephannv: This will make the test helper depend directly on rails. I changed the
```ruby
rendered = controller.view_context.render(view, &block)
Nokogiri::HTML.fragment(rendered)
```

To:
```ruby
view_context = controller&.view_context # controller can be nil if ActionView::TestCase::TestController is not defined
rendered = view.call(view_context: view_context, &block)
Nokogiri::HTML.fragment(rendered)
```

I tested the new code with:
- Rails + RSpec
- Rails + Minitest 
- Ruby Gem (No Rails) + RSpec
- Ruby Gem (No Rails) + Minitest

And it's working fine. I'm finishing the docs to push the changes. If we're not happy with `render_view`, we can change to another name, eg. `render_inline`.
What do you think?

joeldrapper: This is nice.

joeldrapper:
Shipping this as it's a really great start. We can tweak the documentation later.

stephannv:
> Shipping this as it's a really great start. We can tweak the documentation later.

Today I will open a PR improving some code, with requested changes and more tweaks.

PR #277, opened 2022-10-22
Improve test helpers and related documentation

stephannv:
Following up the #275 (Add `render_view` test helper).

Improving `render_view` helper method:
- Now the `render_view` do not depend on `ActionView::TestCase::TestController`, making it Rails independent. 
- `controller` is memoized
- If application doesn't have Nokogiri, MissingTestDependency will be raised with explanation message.
- More tests was added
- Testing docs has now 3 pages:
  - Getting Started
  - RSpec
  - Minitest
- Minitest page has "Asserting with Capybara" with "Coming Soon" alert, because we need to define a `page` method in our helper to Capybara match against it. This will be done in another PR.


joeldrapper (approved):

davidalejandroaguilar:
Where did all of these docs go?

docs/pages/testing/rspec.rb
```diff
@@ -0,0 +1,114 @@
...
+
+						```ruby
+						it "renders items" do
+							result = render_view(Nav.new) do |nav|
+								nav.item(href: "/products/new") { "New Product" }
+								nav.item(href: "/categories/new") { "New Category" }
+							end
+
+							expect(result).to have_link("New Product", href: "/products/new")
+						end
```
davidalejandroaguilar, line 107: @stephannv Could you explain a bit more how this works? The `Nav` class is not defined anywhere on this PR and this looks like a cool usage, but I'm not sure what the `nav.item` methods are doing. Also, this documentation can no longer be found on the website.

stephannv: @davidalejandroaguilar this is just an example, but you could do this:

```ruby
class Nav < Phlex::HTML
  def template(&)
    ul class: "x y z" do
      yield_content(&)
    end
  end
  
  def item(href:, &)
    li class: "a b c" do
      a href: href do
        yield_content(&)
      end
    end
  end
end

render Nav.new do |nav|
  nav.item(href: "/example") { "Example" }
end
```

You can check this page (look all examples tabs too):
https://www.phlex.fun/html/components/

PR #279, opened 2022-10-22
Add spec/test generator

stephannv:
I don't know if this is wanted, but as testing code is common for Ruby developers, I think this will be useful. If application is using RSpec, the `_spec.rb` file will be generated, if application is using Test::Unit will be generated `_test.rb` (minitest compatible). In the future this can be disabled by config, or allow user defines a custom test folder.


joeldrapper:
Generating tests is a good idea, but I think it might be more consistent with Rails if we generate tests with the view generators. What do you think?

stephannv:
> Generating tests is a good idea, but I think it might be more consistent with Rails if we generate tests with the view generators. What do you think?

What do you mean with "more consistent with Rails"? Do you mean about test file content or is it something else?

joeldrapper:
Sorry, I mean all the built-in Rails generators, such as the model generator, also generate the tests. They don't have dedicated test generators.

stephannv:
> Sorry, I mean all the built-in Rails generators, such as the model generator, also generate the tests. They don't have dedicated test generators.

I got it. The `hook_for :test_framework` will call the test generator according to the application test framework. If application is using spec:
```shell
bundle exec rails g phlex:views Nav

create  app/views/nav.rb
  invoke  rspec
  create    spec/views/some_spec.rb
```

If application is using test_unit:
```shell
bundle exec rails g phlex:views Nav

create  app/views/nav.rb
  invoke  test_unit
  create    test/views/nav_test.rb
```

The other option is remove the generators put these test template files inside view generator templates folder and add a test framework check to know which template file should be generated.

stephannv:
I found a bug with RSpec generator, I will investigate this.

stephannv:
@joeldrapper what are your plan for this PR?

joeldrapper:
The PR was held up by all the changes to the testing module names, etc. I’m quite happy with those modules now so we can update this generator to use them. Happy if you want to work on this, otherwise I can update it. Let me know. ❤️

joeldrapper:
Thanks! this looks excellent.

PR #280, opened 2022-10-23
feat: Dynamically defined Rails helpers

joelmoss:
There are quite a few regularly used Rails helpers, and having to call `helpers.*` or to delegate them, as well as having to call `raw` every time can get  a little cumbersome.

This PR replaces the current Rails::Helpers modules with dynamically defined helper method, that wrap the above. Allowing me to simply include `Phlex::Rails::Helpers` into my views.

I've only defined the existing layout helpers, along with `link_to` and `mail_to` for now. Would need to decide whether to implement every Rails helper, or perhaps break them down into groups.

If we implement them all, then including this module will include all helper methods, which I'm not sure we would want. So perhaps creating a module that corresponds to each of the `ActionView::Helpers::*` Rails modules would be a good idea.

Anyway, this new implementation makes it really easy, and for more maintainable.

- [ ] Remove `Phlex::Rails::Layout`
- [ ] Update/create tests
- [ ] Update docs

### DIFF
--- a/fixtures/dummy/app/views/articles/form.rb
+++ b/fixtures/dummy/app/views/articles/form.rb
@@ -3,7 +3,7 @@
 module Views
 	module Articles
 		class Form < Phlex::View
-			include Phlex::Rails::Helpers::FormWith
+			include Phlex::Rails::Helpers
 
 			def template
 				form_with url: "test" do |f|
--- a/fixtures/dummy/app/views/articles/show.html.erb
+++ b/fixtures/dummy/app/views/articles/show.html.erb
@@ -1,0 +1,1 @@
+<%= render Views::LinkHome.new %>
--- a/fixtures/dummy/app/views/articles/show_with_image.html.erb
+++ b/fixtures/dummy/app/views/articles/show_with_image.html.erb
@@ -1,0 +1,1 @@
+<%= render Views::LinkHomeWithImage.new %>
--- a/fixtures/dummy/app/views/link_home.rb
+++ b/fixtures/dummy/app/views/link_home.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Views
+	class LinkHome < Phlex::View
+		include Phlex::Rails::Helpers
+
+		def template
+			link_to "Go Back Home", :root
+		end
+	end
+end
--- a/fixtures/dummy/app/views/link_home_with_image.rb
+++ b/fixtures/dummy/app/views/link_home_with_image.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Views
+	class LinkHomeWithImage < Phlex::View
+		include Phlex::Rails::Helpers
+
+		def template
+			link_to :root, class: "link" do
+				text "Go Back Home"
+				img src: "/go/back.gif"
+			end
+		end
+	end
+end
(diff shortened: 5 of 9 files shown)


joeldrapper:
Hey @joelmoss, thanks for the PR. I really appreciate all your contributions to this project. ❤️ 

I do see there's a fair bit of repetition in the Rails helpers adapters at the moment, but there are a few reasons I’d like to keep them as they are:

1. Not all helpers have the same method signature. Capturing `*args` and `**kwargs` is expensive (performance wise). We don't need to do this because we can just look at the Rails API documentation to see the exact method signatures required for each specific helper method. We also shouldn't need to compute the `block_given?` conditional on helpers that aren't expected to receive blocks. I know this means nasty, ugly repetition for us, but it adds up to performance gains for the user at the end of the day.

2. Not all helpers should be rendered as raw and those that are today might not be tomorrow. As an example, I would like to update many of these tag helpers to use Phlex tags behind the scenes, mirrorign the output of the equivalent Rails helpers. When they have different implementations, it's easy to update them independently.

3. I don't want all the helpers to be included by default because they're not all useful in all contexts. The pattern I’m following is `Phlex::Rails::Helpers::MethodName` so it's really easy to remember how to include exactly what you need. But we can also provide groups of helpers, such as the `Rails::Layout` module which includes all the meta tags for use in layouts.

Hope this all makes sense.

joelmoss:
Thanks for the comments. Just a few of my thoughts in reply...

> ... We don't need to do this because we can just look at the Rails API documentation to see the exact method signatures required for each specific helper method.

The problem with this, is that this will break if any of the method signatures change. So will require monitoring every helper that is reimplemented in Phlex.

> I would like to update many of these tag helpers to use Phlex tags behind the scenes...

I thought of the same thing, but this would end up being a lot of work, especially if the goal is feature parity with the equivalent Rails helper.

joeldrapper:
These are both good points, but I think it's worth the work. You can always access the raw unmodified helpers through the `helpers` proxy, but here we get a chance to adapt them to be better or more suited for Phlex.

I’m also okay with Phlex-Rails-Helpers having feature parity with Rails with a different interface. As an example, if we ever implement `link_to`, it will take a block, not a String so you can put either text or markup inside the link. Additionally, you'll be able to pass it attributes, e.g. 

```ruby
link_to(@article, data: { controller: "Links" }) { @article.title }
```

If we just delegated to Rails `link_to`, this wouldn't work. It's not a big deal at the moment, but it might impact future work. Your proposal #264 for example is incompatible with Rails tag helpers.

joelmoss:
Just out of interest, but why did you decide to require a block when passing content to a tag, and not allow string content as the first arg?

And FYI, Rails link_to helper does accept content as a block.

joeldrapper:
> why did you decide to require a block when passing content to a tag, and not allow string content as the first arg?

We used to have this option but removed it a few weeks ago.

1. It's easier to explain. There's one way to pass content, not two.
2. Content is always in the same place, at the end. Not sometimes at the beginning, sometimes at the end.
3. It's consistent with how you'd pass content when rendering another view. We don't control the initialiser, so we can't allow an optional positional content argument for views.
4. We can always add it back in as a non-breaking-change, but removing it would be a breaking change.

Issue #259, opened 2022-10-14
Callbacks


joelmoss:
Would be great if there were a way to call something before and/or after rendering. Happy to create a PR if you agree.

joeldrapper:
What's the use-case out of interest? Here's one pattern you can use already.

```ruby
class AbstractView < Phlex::View
  def template(&)
    do_somethign
    yield_content(&)
    do_something(&)
  end
end

class ConcreteView < Phlex::View
  def template
    super do
      # things here
    end
  end
end
```

joelmoss:
Yeah I want to avoid having to remember additional code within the `#template` method. It's easy to forget the `super`. As in most cases I will want to before callback to run on all views. I still want to write:

```ruby
def template
  h1 'my view'
end
```

My primary use case is side loading JS and CSS assets along side each view, which simply needs to make a call immediately before `template` is run.

joeldrapper:
How are you thinking of implementing this? I don't want it to impact performance too much, which is quite difficult at this point.

I was originally thinking the best-case performance impact will be one additional method call. We wrap the `template` method and call a different method from `call` instead.

```ruby
def around_template
  template
end
```

But I think we can do even better than this when no callbacks are defined for a view. If we define `around_template` as a `template` alias, it will be as fast as calling the original. Then we only define `around_template` for views that actually have callbacks.

joelmoss:
Already implemented at https://github.com/joelmoss/proscenium/blob/master/lib/proscenium/view_component.rb#L12-L14 and https://github.com/joelmoss/proscenium/blob/master/lib/proscenium/side_load.rb.

This won't affect perf, as all it does is track side loaded asset paths, and render `<script>` and `<link>` tags in my layout when needed.

`around_template` would work, but would fall down when subclassing further, and multiple callbacks are needed/registered. I suppose something akin to ActiveSupport::Callbacks would be ideal.

joeldrapper:
I think it should be okay as long as you always call `super` from `around_template`. But yeah, `around_template` could have an implementation when it's defined to execute a list of before and after callbacks.

joelmoss:
> I think it should be okay as long as you always call super from around_template

Yeah that's reintroducing the problem we sought we solve though. It's extremely easy to forget `super` 😏

joeldrapper:
I think we could do something like this in `Phlex::View#call`.

```ruby
before_render if respond_to?(:before_render)
template(&block)
after_render if respond_to?(:after_render)
```

[...]

PR #294, opened 2022-10-26
feat: Before/after callbacks #259

joelmoss:
Simply prepend `Phlex::View::Callbacks` to any view, and define `#before_template` and/or `#after_template`.

### DIFF
--- a/lib/phlex/view/callbacks.rb
+++ b/lib/phlex/view/callbacks.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Phlex::View::Callbacks
+	def template(&block)
+		respond_to?(:before_template) && before_template
+
+		super(&block)
+
+		respond_to?(:after_template) && after_template
+	end
+end
--- a/lib/phlex/view/callbacks.rb
+++ b/lib/phlex/view/callbacks.rb
@@ -2,10 +2,10 @@
 
 module Phlex::View::Callbacks
 	def template(&block)
-		respond_to?(:before_template) && before_template
+		before_rendering_template if respond_to?(:before_rendering_template)
 
 		super(&block)
 
-		respond_to?(:after_template) && after_template
+		after_rendering_template if respond_to?(:after_rendering_template)
 	end
 end
--- a/test/phlex/view/callbacks.rb
+++ b/test/phlex/view/callbacks.rb
@@ -9,7 +9,7 @@ describe Phlex::View do
 		view do
 			prepend Phlex::View::Callbacks
 
-			def before_template
+			def before_rendering_template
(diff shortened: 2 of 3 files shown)


lib/phlex/view/callbacks.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Phlex::View::Callbacks
+	def template(&block)
+		respond_to?(:before_template) && before_template
```
joeldrapper, line 5: I agree it shouldn't be `before_render`, since we have a `render` method but can we go with `before_rendering`? `template` is a noun and we can't do something before a noun. `before_rendering` would be short for `before_rendering_template`.

lib/phlex/view/callbacks.rb
```diff
@@ -0,0 +1,11 @@
...
+
+		super(&block)
+
+		respond_to?(:after_template) && after_template
```
joeldrapper, line 9: Nit: I think these might read better written like this.

```suggestion
		after_template if respond_to?(:after_template)
```

joeldrapper (commented):
I left a comment about the name of the callback methods. Implementation is perfect. Looking forward to getting this merged in!

test/phlex/view/callbacks.rb
```diff
@@ -0,0 +1,29 @@
...
+			end
+
+			def after_template
+				h3 { "Bye" }
+			end
+		end
+
+		it "calls before and after template" do
+			expect(output).to be == "<h1>Hello</h1><h2>World</h2><h3>Bye</h3>"
+		end
```
joeldrapper, line 27: Really simple test. Love it!

lib/phlex/view/callbacks.rb
joelmoss: Not sure doing something before a noun actually matters. I certainly don't think anyone will complain.

However, the important thing is that `before_template` is more accurate and says exactly what it does, whereas `before_rendering` is kinda misleading, and doesn't actually imply it is called before the template.

joeldrapper: I’m happy to go with `before_rendering_template` to be really clear.

joelmoss: Compromise 👌

joelmoss: What about `after_template`? Not sure "rendering" should be used for that one, as it's not actually called after rendering.

joeldrapper: It’s executed in the `call` method which is essentially a "render_template" method.

joelmoss:
All done!

Issue #297, opened 2022-10-26
Update the `_attributes` method to take keyword arguments


joeldrapper:
The `fetch_attrs` method won't be necessary if the `_attributes` method takes keyword arguments instead of a positional hash argument.

joelmoss:
So I'm a little concerned that after removing the `fetch_attrs` method, all tests still pass 🤔 including those which still call that method.

I'm not familiar enough with the compilation stuff to be confident that all still works, especially considering all tests still pass, even though they are calling a method that no longer exists.

joeldrapper:
Oh, that's because the tests test that it compiles to call `fetch_attrs` but none of the tests actually execute the compiled code just yet. You should be able to do a find / replace for `fetch_attrs` -> `process_attributes`.

joeldrapper:
Compilation is experimental at this point, so doesn't matter if it breaks.

PR #299, opened 2022-10-26
feat: Rename _attributes to process_attributes (#297 and #264)

joelmoss:


lib/phlex/view.rb
```diff
@@ -182,11 +182,7 @@ def helpers
 			@_view_context
 		end
 
-		def fetch_attrs(**kwargs)
-			Phlex::ATTRIBUTE_CACHE[kwargs.hash] || _attributes(kwargs)
-		end
-
-		def _attributes(attributes, buffer: +"")
+		def process_attributes(buffer: +"", **attributes)
```
joeldrapper, line 185: Let's change the `buffer` to a positional here so it can't be overridden by a `buffer` attribute. I think this will mean we need to update calls from `Phlex::HTML` to put `**` before the attributes they pass in.

```suggestion
		def process_attributes(buffer = +"", **attributes)
```

PR #312, opened 2022-11-01
Remove iteration details from collections

joeldrapper:
It was fun to experiment with these, but I’m of quite a strong opinion that they should be implemented using CSS counters and nth selectors instead. This way, state doesn't need to be passed in when rendering an item.

### DIFF
--- a/docs/pages/library/collections.rb
+++ b/docs/pages/library/collections.rb
@@ -32,16 +32,6 @@ module Pages
 						end
 						```
 
-						You can also access the predicates `first?` and `last?` and the instance variables `@index`, `@position` and `@collection`.
-
-						```ruby
-						def item_template
-							li **classes(first?: "border-t") do
-								"\#{@position}/\#{@collection.size} \#{@item}"
-							end
-						end
-						```
-
 						## Rendering a collection
 
 						Putting it all together, we can create a `List` view that renders each item in an `<li>`, all wrapped up in an outer `<ul>`.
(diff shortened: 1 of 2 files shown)


dhnaranjo:
I was using this to support creating a dynamically generated list of radio groups that work with `accepts_nested_attributes`, like `things_attributes[@index][option_id]`. Think like a user-created survey. The immediate result is that all group names end up `things_attributes[][option_id]`. I can just replace it with literally anything else so long as it's unique and things work fine, but that's a non-style-related use of the index value, and I did prefer that to what I'm doing now.

dhnaranjo:
I was using this to populate a strong params acceptable list of inputs, `things_attributes[@index][option_id]`, think a user-defined survey. 

The fix after this change is to just use a different unique identifier like a record ID. Trouble is this strong params `params.require(:form).permit(things_attributes: [:option_id]` works fine with integers but not with string keys, dunno why, but UUIDs are out. In my case I did something like `@item.id.gsub("-","").hex`.

It works, but I definitely prefer how it was. I get the abuse you wanna avoid with folks adding CSS classes to specific elements, table striping and such, instead of using appropriate selectors, but I bet there's a fair number of reasons people would want to be able to work with these index values.

joeldrapper:
Hey @dhnaranjo thanks for the use-case! I don’t quite follow your example with strong parameters but I wonder if it might be something we could solve with the upcoming phlex-rails form builder we're working on. Could you share some more details about the problem in a discussion thread?

dhnaranjo:
Done @ https://github.com/joeldrapper/phlex/discussions/373

PR #318, opened 2022-11-02
Remove `process_attributes` feature

joeldrapper:
@joelmoss I’m really sorry but I have to remove this feature for now as it's broken by the caching. I tried to change attribute caching to be per-class, but the performance is so much worse because we have to use `self.class.attribute_cache` or `Phlex::ATTRIBUTE_CACHE[self.class.name]` and both require multiple method calls. I’ve got some ideas about how we might be able to put it back in with a mixin, but it's got to go for now.

It would be quite easy to support a global attribute processing function so that might be an option, e.g.

```ruby
Phlex.configure do |config|
  config.process_attributes do |attributes|
    bla
  end
end
```

### DIFF
--- a/docs/pages/templates.rb
+++ b/docs/pages/templates.rb
@@ -77,32 +77,6 @@ module Pages
 				end
 
 				render Markdown.new(<<~MD)
-					## Processing Attributes
-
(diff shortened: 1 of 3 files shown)


joelmoss:
OK, no problem. Thx anyway. Will think on this

PR #322, opened 2022-11-02
[Proposal] Add render? method to allow conditional rendering

stephannv:
The idea is the same as this one:
https://viewcomponent.org/guide/conditional_rendering.html

I know that with phlex you can do something like:
```ruby
def template
  return if my_method?
  
  ...
end
```

But with built-in support, we can ensure consistency between components and encourage a framework pattern. What do you think?

My use case:
```ruby
class Toast < ApplicationView
  def initialize(flash)
    @type, @message = flash.first
  end
  
  private
  
  attr_reader :type, :message
  
  def template
    ... # template code here
  end
  
  def render?
    message.present?
  end
end
```

I solved this problem in my project with:
```ruby
class ApplicationView < Phlex::View
  def call(...)
    return unless render?
    
    super(...)
  end
  
  private
  
  def render?
    true
  end
end
```

### DIFF
--- a/lib/phlex/view.rb
+++ b/lib/phlex/view.rb
@@ -29,6 +29,8 @@ module Phlex
 		end
 
 		def call(buffer = +"", view_context: nil, parent: nil, &block)
+			return buffer unless render?
+
 			raise "The same view instance shouldn't be rendered twice" if rendered?
 
 			@_rendered = true
@@ -46,6 +48,10 @@ module Phlex
 
 		def rendered?
 			@_rendered ||= false
+		end
+
+		def render?
+			true
 		end
 
 		HTML::STANDARD_ELEMENTS.each do |method_name, tag|
--- a/test/phlex/view/call.rb
+++ b/test/phlex/view/call.rb
@@ -23,4 +23,21 @@ describe Phlex::View do
 			end
 		end
 	end
+
+	with "`render?` method returning false" do
+		view do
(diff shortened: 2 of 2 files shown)


joeldrapper:
Based on my benchmarks for #259, my guess is this will add a ~1-3% performance overhead even when the feature is unused. I wonder if we could make it opt-in with module inclusion, like how callbacks were implemented.

joeldrapper:
Actually, it doesn't look like it's that significant. I wonder if YJIT will inline stuff like this eventually anyway.

evenreven:
My two cents from the sidelines... If the overhead is negligible, I for one would love this. I've used this with view_component in the past, and I found it really useful. I had to use a lot of conditional rendering in an app, and just being able to put the conditional logic in the component itself in a basically self-documenting `render?` method is so much better than the if-o-rama my templates were turning into. One of my favourite features of view_component, easily.

joeldrapper:
That’s weird. I think we should ship this once you've removed the benchmark change. We should definitely return the `buffer` instead of `nil` too.

stephannv:
> That’s weird. I think we should ship this once you've removed the benchmark change. We should definitely return the `buffer` instead of `nil` too.

Ok, I'm removing the benchmark code now and returning `buffer` instead nil.

joeldrapper (approved):

PR #324, opened 2022-11-08
Generator task for Rails controller with Phlex views

barrywoolgar:
As mentioned [here](https://github.com/joeldrapper/phlex/discussions/235#discussioncomment-3852112), it seems desirable to be able to emulate the Rails controller generator:

`rails generate controller Home index about`

This PR provides a single generator that creates a Rails controller with routes, actions, and the set of Phlex views. Each action renders the view, and the default content of the view matches that of the default ERB template for an action:

`rails generate phlex:controller Home index about`

This generates:

```ruby
# app/controllers/home_controller.rb
class HomeController < ApplicationController
  def index
    render Views::Nice::Index.new
  end

  def about
    render Views::Nice::About.new
  end
end

# app/views/home/index.rb - if app/views/layout.rb exists
module Views
  class Home::Index < Phlex::HTML
    include ApplicationView

    def template
      render Layout.new(title: "Home - Index") do
        h1 { "Home#index" }
        p { "Find me in app/views/home/index.rb" }
      end
    end
  end
end

# app/views/home/about.rb - if app/views/layout.rb does not exist
module Views
  class Home::About < Phlex::HTML
    include ApplicationView

    def template
      # render Layout.new(title: "Home - About") do
        h1 { "Home#about" }
        p { "Find me in app/views/home/about.rb" }
      # end
    end
  end
end

```

I hope this is useful, please let me know if there's anything that could be improved. Did you want a commit updating the docs too?

If #279 gets merged I'll have to come back and check that view file tests are created alongside the controller.

(I ran the tests btw, but I'm getting the same two failures before and after these changes)

### DIFF
--- a/lib/generators/phlex/controller/USAGE
+++ b/lib/generators/phlex/controller/USAGE
@@ -2,9 +2,9 @@ Description:
   Generates a Rails controller with Phlex views for the supplied actions
 
 Example:
-  rails generate phlex:controller Home index show
+  rails generate phlex:controller Articles index show
 
 This will create:
-  app/controllers/home.rb
-  app/views/home/index.rb
-  app/views/home/show.rb
+  app/controllers/articles_controller.rb
+  app/views/articles/index.rb
+  app/views/articles/show.rb
--- a/lib/generators/phlex/controller/USAGE
+++ b/lib/generators/phlex/controller/USAGE
@@ -1,0 +1,10 @@
+Description:
+  Generates a Rails controller with Phlex views for the supplied actions
+
+Example:
+  rails generate phlex:controller Home index show
+
+This will create:
+  app/controllers/home.rb
+  app/views/home/index.rb
+  app/views/home/show.rb
--- a/lib/generators/phlex/controller/controller_generator.rb
+++ b/lib/generators/phlex/controller/controller_generator.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module Phlex
+	module Generators
+		class ControllerGenerator < ::Rails::Generators::NamedBase # :nodoc:
+			source_root File.expand_path("templates", __dir__)
+
+			argument :actions, type: :array, default: [], banner: "action action"
+			class_option :skip_routes, type: :boolean, desc: "Don't add routes to config/routes.rb."
+			class_option :parent, type: :string, default: "ApplicationController", desc: "The parent class for the generated controller"
+
+			check_class_collision suffix: "Controller"
+
+			def create_controller_files
+				template "controller.rb.erb", File.join("app/controllers", class_path, "#{file_name}_controller.rb")
+			end
+
+			def copy_view_files
+				base_path = File.join("app/views", class_path, file_name)
+				empty_directory base_path
+
+				actions.each do |action|
+					Rails::Generators.invoke("phlex:view", [remove_possible_suffix(name) + "/" + action])
+				end
+			end
+
+			def add_routes
+				return if options[:skip_routes]
+				return if actions.empty?
+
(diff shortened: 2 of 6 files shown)


joeldrapper:
@barrywoolgar this is great! Should we use the `phlex:page` generator for the views?

barrywoolgar:
Yes, that would be a lot better than me half-duplicating the functionality - though I do think telling people where to find the view file is helpful. I'll have a look this evening at calling that code instead.

joeldrapper:
Thanks. I believe the test errors are unrelated to your change and caused by a gem update. I haven't tracked down which one yet. Maybe we should commit a `Gemfile.lock`.

barrywoolgar:
I've sorted out the Rubocop errors, and had a look at using `phlex:page`. I feel like I should be able to define a `hook_for` somewhere but I'm not quite seeing it yet - will keep looking.

It seems like the `phlex:controller` code gets run even with a plain `rails g controller`, but that will trying to generate ERB templates unless you override the `template_engine` generator config to `:phlex`. 

Is your preference for people to use the `phlex:` commands over the standard Rails generators? 

Also, I've added in the Layout to the view files, but there's nothing to guarantee this class will exist. Do you think the `phlex:install` command should create a default `Layout` class file too?

joeldrapper:
> It seems like the phlex:controller code gets run even with a plain rails g controller, but that will trying to generate ERB templates unless you override the template_engine generator config to :phlex.

Phlex isn't really an ActionView template engine — it's an ActionView replacement, so I don't think we can do that. I’m not sure the best way to stop Rails from generating ERB views but it might be best to keep the genreator namespaced under `phlex:` for now.

> Also, I've added in the Layout to the view files, but there's nothing to guarantee this class will exist. Do you think the phlex:install command should create a default Layout class file too?

That's complicated. If you've installed Phlex in a new project, you'll probably want to generate a layout and use that from your page views. But if you're bringing it into a new project, you'll probably want to keep your ERB layout and use the `phlex:view` generator instead. (see https://www.phlex.fun/rails/migrating/)

barrywoolgar:
> Should we use the phlex:page generator for the action views since that will include the layout?

I figured out how to do that now, but I've used `phlex:view` instead - I'm not sure of the intended use for the various `phlex:*` generators is, but it feels more Rails-conventional to use `phlex:view` in a controller context.

> it might be best to keep the generator namespaced under phlex: for now

Agreed. I think I had some "spare" code in my test app that was making me think this Phlex controller generator was running on a normal `rails g controller`. Sorry for the confusion.

> That's complicated.

Good point about Layout being optional. I've added a naive check for `app/views/layout.rb`, but ideally we'd look for a view file that has `Phlex::Rails::Layout` as an included module. I wasn't able to get that working though, perhaps due to lazy class loading in development. If a layout is found it uses the wrapped markup, and if not the example Layout wrapper is commented out as example code.

I think the USAGE files are bit out of sync, but I'll update them once you're happy the code is working as desired.

lib/generators/phlex/controller/USAGE
```diff
@@ -0,0 +1,10 @@
+Description:
+  Generates a Rails controller with Phlex views for the supplied actions
+
+Example:
+  rails generate phlex:controller Home index show
+
+This will create:
+  app/controllers/home.rb
+  app/views/home/index.rb
+  app/views/home/show.rb
```
joeldrapper, line 10:
```suggestion
  rails generate phlex:controller Articles index show

This will create:
  app/controllers/articles_controller.rb
  app/views/articles/index.rb
  app/views/articles/show.rb
```

joeldrapper:
I updated the USAGE file to use `articles` as the example becuase I think the more common use-case is a plural resource.

PR #328, opened 2022-11-13
Add support to custom form builders

stephannv:
Related to https://github.com/joeldrapper/phlex/issues/327


joeldrapper:
Thanks @stephannv! We might revisit this but hopefully we’ll replace `form_with` with `Phlex::Rails::Form` soon.

stephannv:
> Thanks @stephannv! We might revisit this but hopefully we’ll replace `form_with` with `Phlex::Rails::Form` soon.

No problem.

Don't you think it's wise to keep rails form_with support? I mean, with Phlex::Rails::Form you will have to keep a mirror from native Rails FormBuilder features, and not everyone will want/be able to abandon the rails form builder. Or do you intend to keep support for both (Default form builder + Phlex::Rails::Form) without supporting custom rails form builders?

joeldrapper:
> Or do you intend to keep support for both (Default form builder + Phlex::Rails::Form) without supporting custom rails form builders?

This for now. ^^

PR #330, opened 2022-11-13
Whitespace around markdown `code`

joeldrapper:


lib/phlex/html.rb
```diff
@@ -216,6 +216,12 @@ def _output(content)
 
 		def whitespace
 			@_target << " "
+
+			if block_given?
+				yield
+				@_target << " "
+			end
```
joeldrapper, line 223: This allows us to wrap things in whitespace.

PR #341, opened 2022-11-16
Allow `Phlex::Turbo::Stream` element to take arbitrary attributes

marcoroth:
With the Introduction of Custom Turbo Stream Actions it's not mandatory to always just pass the `action` and `target` attributes on the `<turbo-stream>` element. 

Since you can write actions which don't target any elements you also don't need to specify the `target` or `targets` attribute (right now with Phlex `0.5` you also wouldn't be able to pass over a `targets` attribute).

Additionally, you need to be able to pass any attribute to the `<turbo-stream>` element so you can use custom attributes in your custom action.

For more context: https://github.com/hotwired/turbo-rails/pull/375 and https://github.com/hotwired/turbo-rails/pull/373

### DIFF
--- a/lib/phlex/turbo/stream.rb
+++ b/lib/phlex/turbo/stream.rb
@@ -5,13 +5,13 @@ module Phlex
 		class Stream < Phlex::HTML
 			register_element :turbo_stream
 
-			def initialize(action:, target:)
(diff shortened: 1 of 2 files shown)


joeldrapper:
Makes sense. 👍

PR #369, opened 2022-11-21
Use `around_template` hook for callbacks

joeldrapper:
Switch to using the `around_template` method for `Phlex::Callbacks`. This means we don't need to use `prepend` and patch the main `template` method. We can also define the `before_rendering_template` and `after_rendering_template` to return `nil` by default and just call them rather than checking if they’re defined first, saving time.

### DIFF
--- a/lib/phlex/html/callbacks.rb
+++ b/lib/phlex/html/callbacks.rb
@@ -1,11 +1,21 @@
 # frozen_string_literal: true
 
 module Phlex::HTML::Callbacks
-	def template(&block)
-		before_rendering_template if respond_to?(:before_rendering_template)
+	def self.prepended(mod)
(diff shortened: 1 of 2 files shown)


lib/phlex/html/callbacks.rb
```diff
@@ -1,11 +1,21 @@
 # frozen_string_literal: true
 
 module Phlex::HTML::Callbacks
-	def template(&block)
-		before_rendering_template if respond_to?(:before_rendering_template)
+	def self.prepended(mod)
+		raise "Phlex::HTML::Callbacks should be included rather than prepended."
+	end
```
joeldrapper, line 6: Temporary "deprecation" warning in case anyone is prepending `Phlex::Callbacks`.

PR #398, opened 2022-11-27
`format_object` method

joeldrapper:
A simpler take on #397, allows you to define object formatters like this.

```ruby
def format_object(object)
  case object
  when Date
    object.strftime("%d/%m/%y")
  when Money
    object.format
  else
    super
  end
end
```

The interface isn’t as nice, but:

1. it gets inheritance right when you call `super` in the `else` condition;
2. you can easily define `format_object` in a module, such as `ApplicationView`;
3. it's fast; and
4. you can use case equality or pattern matching.

### DIFF
--- a/lib/phlex/html.rb
+++ b/lib/phlex/html.rb
@@ -179,7 +179,8 @@ module Phlex
 				case content
 					when String then content
 					when Symbol then content.name
-					else content.to_s
+					when Integer, Float then content.to_s
+					else format_object(content) || content.to_s
 				end
 			)
 
@@ -234,6 +235,10 @@ module Phlex
 			true
 		end
 
+		private def format_object(oject)
+			nil
+		end
+
 		private def around_template
 			before_template
 			yield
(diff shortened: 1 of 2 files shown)


lib/phlex/html.rb
```diff
@@ -187,7 +187,8 @@ def text(content)
 				case content
 					when String then content
 					when Symbol then content.name
-					else content.to_s
+					when Integer, Float then content.to_s
```
joeldrapper, line 190: I think it makes sense to have a fast-path for Integer, Float, Symbol and String since these are the most common objects and it would be unusual to want to format them in any other way than this.

lib/phlex/html.rb
```diff
@@ -187,7 +187,8 @@ def text(content)
...
+					else format_object(content) || content.to_s
```
joeldrapper, line 191: We fall back to `content.to_s` here because this object has been intentionally passed to `text` meaning we know the developer wants to output something. We don’t do this in `yield_content` becuase that would result in potentially calling `to_s` on an empty Enumeratorand outputting an Enumerator object description.

francois (commented):
Personally, I prefer this method to #397, as it is pure Ruby semantics.

PR #418, opened 2023-01-07
Render data hash with string keys

djfpaagman:
I was building a component with some Stimulus `data-` attributes and noticed you can't use string keys for the `data` hash, for example:

```ruby
def template
	div data: { "name_first_name" => "Joel" }
end
```

will give:

```
describe Phlex::HTML with string keyed hash attributes it dasherizes the attributes test/phlex/view/attributes.rb:25
        ⚠ NoMethodError: undefined method `name' for "name_first_name":String
                /Users/dennis/Code/phlex/lib/phlex/html.rb:357 block (2 levels) in _build_attributes
```

I'm opening this PR as a starting point. I think using `to_s` has some downsides as it makes a copy of the string/symbol every time so memory wise it's not a great solution. But i'm also not sure if a class check is worth it either, probably most of the times a hash with symbol keys will be used. 

What do you think?

### DIFF
--- a/lib/phlex/html.rb
+++ b/lib/phlex/html.rb
@@ -354,7 +354,7 @@ module Phlex
 				when Symbol
 					buffer << " " << name << '="' << ERB::Util.html_escape(v.name) << '"'
 				when Hash
-					_build_attributes(v.transform_keys { "#{k}-#{_1.name.tr('_', '-')}" }, buffer: buffer)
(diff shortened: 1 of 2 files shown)


joeldrapper:
I accidentally merged the other PR from the GitHub iOS app. I reverted that and opened a new PR that reverts the revert so we can discuss it here: https://github.com/joeldrapper/phlex/pull/420

PR #420, opened 2023-01-07
Render data hash with string keys, take 2

joeldrapper:
Reverts joeldrapper/phlex#419, which reverted #418 which I accidentally merged. I’m copying below the original description from #418, since this PR is a replacement for that:

from @djfpaagman

I was building a component with some Stimulus `data-` attributes and noticed you can't use string keys for the `data` hash, for example:

```ruby
def template
	div data: { "name_first_name" => "Joel" }
end
```

will give:

```
describe Phlex::HTML with string keyed hash attributes it dasherizes the attributes test/phlex/view/attributes.rb:25
        ⚠ NoMethodError: undefined method `name' for "name_first_name":String
                /Users/dennis/Code/phlex/lib/phlex/html.rb:357 block (2 levels) in _build_attributes
```

I'm opening this PR as a starting point. I think using `to_s` has some downsides as it makes a copy of the string/symbol every time so memory wise it's not a great solution. But i'm also not sure if a class check is worth it either, probably most of the times a hash with symbol keys will be used. 

What do you think?

### DIFF
--- a/lib/phlex/html.rb
+++ b/lib/phlex/html.rb
@@ -354,7 +354,7 @@ module Phlex
 				when Symbol
 					buffer << " " << name << '="' << ERB::Util.html_escape(v.name) << '"'
 				when Hash
-					_build_attributes(v.transform_keys { "#{k}-#{_1.name.tr('_', '-')}" }, buffer: buffer)
+					_build_attributes(v.transform_keys { "#{k}-#{_1.to_s.tr('_', '-')}" }, buffer: buffer)
 				else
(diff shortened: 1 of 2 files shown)


joeldrapper:
@djfpaagman I think Phlex should support nested string keys, but this implementation isn't quite right. We shouldn't dasherize keys when they are already strings that way string keys provide a way to use real underscores. If you want dasherization, but need to use other characters like `:` (for Alpine), you can alway use a special symbol literal, e.g. `:"my_very:special_key"`.

joeldrapper:
Great point, I just opened an issue for that here. https://github.com/joeldrapper/phlex.fun/issues/11

PR #427, opened 2023-01-10
Add experimental `DeferredRender` module

joeldrapper:
`DeferredRender` yields the content block early, throwing away the results, so you can render your template having already captured everything with your own DSL.

### DIFF
--- a/lib/phlex/deferred_render.rb
+++ b/lib/phlex/deferred_render.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Phlex
+	module DeferredRender
+		include Experimental
+
+		def template(&block)
+			capture(&block) # yield the block and throw away the output
+			super() # empty parens ensure we don't pass the block which could be yielded a second time
+		end
+	end
+end


joelmoss:
Out of interest, what would be the use case for this?

joeldrapper:
Let's say you have a tab group component with this interface.

```ruby
render TabGroup do |t|
  t.tab("A") { "Content for A" }
  t.tab("B") { "Content for B" }
  t.tab("C") { "Content for C" }
end
```

You can easily implement it like this.

```ruby
class TabGroup < Phlex::HTML
  def template(&content)
    div(class: "tab-group", &content)
  end
  
  def tab(name, &content)
    a(class: "tab-nav") { name }
    div(class: "tab-content", &content)
  end
end
```

But your tabs are output like this.

```
- name
- content

- name
- content

- name
- content
```

If you needed them to be output like this

```
- name
- name
- name

- content
- content
- content
```

You could use `DeferredRender`.

```ruby
class TabGroup < Phlex::HTML
  Tab = Data.define(:name, :content)
  
  prepend DeferredRender
  
  def initialize
    @tabs = []
  end
  
  def template
    div(class: "tab-group") {
      ul(class: "tab-nav") {
        @tabs.each do |t|
          li {
            t.name
          }
        end
      }
      
      ul(class: "tab-content") {
        @tabs.each do |t|
          li(&t.content)
        end
      }
    }
  end
  
  def tab(name, &content)
    @tabs << Tab.new(name, content)
  end
end
```

bradgessler:
`prepend DeferredRender` feels like it exposes APIs that are internal to the `Phlex::HTML` class.

How about something like this? If a dev wants to implement this, they'd do this:

```ruby
class TabGroup < Phlex::HTML
  Tab = Data.define(:name, :content)
   
  def initialize
    @tabs = []
  end
  
  def deferred_template
    div(class: "tab-group") {
      ul(class: "tab-nav") {
        @tabs.each do |t|
          li { t.name }
        end
      }
      
      ul(class: "tab-content") {
        @tabs.each do |t|
          li(&t.content)
        end
      }
    }
  end
  
  def tab(name, &content)
    @tabs << Tab.new(name, content)
  end
end
```

Then the implementation looks something like this:

```ruby
 module Phlex
  module DeferredRender
    include Experimental

    def template(&block)
      if responds_to? :deferred_template
        capture(&block) # yield the block and throw away the output
        super() # empty parens ensure we don't pass the block which could be yielded a second time
      else
        super(...)
      end
    end
  end
 end
```

I'm not sure I got those right, but the idea is that if a `deferred_template` exists, it defers the rendering. I think there's still something a bit off with the naming, but I can't quite pin it down.

bradgessler:
Another idea, if that doesn't work or make sense, is to try to keep as many things as an `include` as possible. Thus you'd end up with something like this:

```ruby
class TabGroup < Phlex::HTML
  Tab = Data.define(:name, :content)
  
  include DeferredRendering
  
  def initialize
    @tabs = []
  end
  
  def template
    div(class: "tab-group") {
      ul(class: "tab-nav") {
        @tabs.each do |t|
          li { t.name }
        end
      }
      
      ul(class: "tab-content") {
        @tabs.each do |t|
          li(&t.content)
        end
      }
    }
  end
  
  def tab(name, &content)
    @tabs << Tab.new(name, content)
  end
end
```

That include would be as small as this:

```ruby
module DeferredRendering
  def self.included
    prepend DeferredRender 
  end
end
```

As a developer, I ideally don't want to have to think about "do I prepend that or include it?" when working with a library.

joeldrapper:
I tend to agree @bradgessler.

PR #433, opened 2023-01-12
Guard for href as non-string

joelmoss:
The check for JS in `href` attribute fails if `href` is not a string. This PR simply guards against that.

### DIFF
--- a/lib/phlex/html.rb
+++ b/lib/phlex/html.rb
@@ -320,7 +320,7 @@ module Phlex
 		end
 
 		private def __attributes__(**attributes)
-			if attributes[:href]&.start_with?(/\s*javascript/)
+			if attributes[:href].is_a?(String) && attributes[:href].start_with?(/\s*javascript/)
 				attributes[:href] = attributes[:href].sub(/^\s*(javascript:)+/, "")
 			end
 


joeldrapper:
Do you have any examples of a non-String `href`?

joelmoss:
Well I'm actually overriding `Phlex::HTML#_attributes` so I can add additional functionality to certain attributes. In this, I am passing the `href` attribute to Rails `url_for`. But that fails because `url_for` accepts non-strings.

Spone:
`href` could be for instance: `[:admin, :users, User.first]`

joeldrapper:
@joelmoss Overriding `_attributes` is an unsupported private API at the moment, but it should be fine as long as you always call `super` so you get the safety from Phlex. If you’re overriding `_attributes`, can’t you do the `url_for` call in your override and then pass Phlex an already resolved string?

joelmoss:
Yeah I appreciate it's a private API, but it's the best way for me to do what I want.

Sorry, I actually meant to say that I'm overriding `_build_attributes`, not `_attributes`. But looking into it more, I can instead use `_attributes` like you say. I'd got it into my head that `_build_attributes` was a better place, but apparently not 😏

> An alternative pattern is to define your own link helper and use that instead of a.

In this case, I want to avoid doing this, as it requires remembering to use the helper instead of the element. There are also some other attributes I "decorate".

Anyway, using `_attributes` works and doesn't need this PR. So do what you want this. thx

joeldrapper:
@joelmoss that's great! Allowing that to be overridden as `attributes` without the underscore is something I'm thinking about. It's not ideal that you have to call and return `super` though. I know we talked about some other options before.

Another thing you might find useful that we could probably add would be some method to check which element is currently being rendered. That would allow you to target your `href` override to `a` tags only.

joelmoss:
Yeah, I've been thinking about a more built in way to do this...

```ruby
customise_attribute :href, on: :a do |attr_value|
  '/new/href'
end
```

joeldrapper:
@joelmoss at the moment, you can do that like this.

```ruby
def a(href:, **kwargs, &block)
  super(href: "/new/href", **kwargs, &block)
end
```

joelmoss:
Yup. That works if I want to override a specific element. But not if if want to modify the attribute for several or all elements. For example, the class attribute.

PR #438, opened 2023-01-15
Detect block passed to `new`

joeldrapper:
It’s easy to accidentally pass a block to the class method `new` instead of the instance method `render`. For example, if you forget the parentheses around a brace block when rendering a view inside another view.

```ruby
# Block sent to `render`
render(Example.new) { "Hello" }

# Block sent to `new`
render Example.new { "Hello" }
```

We can handle this gracefully and save the block for later by overriding `.new`.

There's a performance cost to this, but I think it's worth it.

Before
```
ruby 3.2.0 (2022-12-25 revision a528908271) [arm64-darwin22]
                Page     9.314k i/100ms
                Page     93.119k (± 0.7%) i/s -    465.700k in   5.001396s
```

After
```
ruby 3.2.0 (2022-12-25 revision a528908271) [arm64-darwin22]
                Page     9.194k i/100ms
                Page     91.961k (± 0.2%) i/s -    468.894k in   5.098888s
```

### DIFF
--- a/lib/phlex/html.rb
+++ b/lib/phlex/html.rb
@@ -137,9 +137,11 @@ module Phlex
 			end
 			alias_method :render, :call
 
-			def new
-				if block_given?
-					raise ArgumentError, "You passed a block to #{name}.new. You probably meant to pass it to #{name}#call."
+			def new(*args, &block)
+				if block
+					object = super(*args, &nil)
+					object.instance_variable_set(:@_content_block, block)
+					object
 				else
 					super
 				end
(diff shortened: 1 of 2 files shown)


lib/phlex/html.rb
```diff
@@ -136,6 +136,14 @@ def call(...)
 				new(...).call
 			end
 			alias_method :render, :call
+
+			def new
+				if block_given?
+					raise ArgumentError, "You passed a block to #{name}.new. You probably meant to pass it to #{name}#call."
```
joeldrapper, line 142: This error message isn’t particularly helpful because most of the time, you should actually pass the block to the `render` method on another view.

lib/phlex/html.rb
```diff
@@ -136,17 +136,29 @@ def call(...)
 				new(...).call
 			end
 			alias_method :render, :call
+
+			def new(*args, &block)
+				if block
+					object = super(*args, &nil)
```
joeldrapper, line 142: We don't pass the block up to `super` because we don't want it to be handled by a users initialiser.

PR #465, opened 2023-01-28
Remove experimental `collection` and `table` abstractions

### LINKED ISSUE #213: Collections pattern
Introduced experimentally in #125. If this is the direction we want to go, it will need tests and documentation.

joeldrapper:
I think we should keep the core library small and remove these experiments. Closes #213


hangsu:
I really like this decision and the sentiment behind it.

A table abstraction is so custom and it's impossible to hit the right level of abstractness vs usefulness. that works for everyone's use-case.

I think a lot of the problems these built-in abstractions solve can be better solved by including interesting examples in the documentation that improves understanding of the core concepts in addition to providing a starting point for patterns.

PR #482, opened 2023-02-09
Add YARD docs for HTML tags

joeldrapper:
image


marcoroth:
It might be interesting to make sure it's compatible with [`sord`](https://github.com/AaronC81/sord). That way we would also get RBS annotations for free!

joeldrapper:
It should be if they support YARD. I would prefer to use RBS, but at the moment it looks like YARD has by far the best text editor support.

marcoroth:
Technically it can convert YARD docs to RBI/RBS annotations

PR #487, opened 2023-02-12
RBS signatures

joeldrapper:


joeldrapper:
I enjoyed experimenting with RBS signatures, but steep isn't quite at the point where it can handle all the Ruby syntax necessary for this to work for us.

PR #490, opened 2023-02-14
Phlex::HTML as a module

joeldrapper:
Just prototyping what Phlex would look like if `Phlex::HTML` were a module.

### DIFF
--- a/fixtures/view_helper.rb
+++ b/fixtures/view_helper.rb
@@ -10,7 +10,7 @@ module ViewHelper
 
 	def view(&block)
 		let :view do
-			Class.new(Phlex::HTML, &block)
+			Class.new { include Phlex::HTML }.tap { |c| c.class_eval(&block) }
 		end
 	end
 end
--- a/fixtures/layout.rb
+++ b/fixtures/layout.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 module Example
-	class LayoutComponent < Phlex::HTML
+	class LayoutComponent
+		include Phlex::HTML
+
 		def initialize(title: "Example")
 			@title = title
 		end
--- a/fixtures/page.rb
+++ b/fixtures/page.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 module Example
-	class Page < Phlex::HTML
+	class Page
+		include Phlex::HTML
+
 		def template
 			render LayoutComponent.new do
 				h1 { "Hi" }
--- a/lib/phlex/html.rb
+++ b/lib/phlex/html.rb
@@ -5,7 +5,7 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("3.0")
 end
 
 module Phlex
-	class HTML
+	module HTML
 		# A list of standard elements that have been registered.
 		STANDARD_ELEMENTS = Concurrent::Map.new
 
@@ -17,10 +17,14 @@ module Phlex
 
 		UNBUFFERED_MUTEX = Mutex.new
 
-		extend Elements
 		include Helpers, VoidElements, StandardElements
 
-		class << self
+		def self.included(base)
+			base.extend(Elements)
+			base.extend(ClassMethods)
+		end
+
+		module ClassMethods
 			# Render the view to a String. Arguments are delegated to <code>new</code>.
 			def call(...)
 				new(...).call
@@ -53,6 +57,14 @@ module Phlex
 						@unbuffered_class = Class.new(Unbuffered)
 					end
 				end
+			end
+
+			def method_added(method_name)
+				if method_name[0] == "_" && Phlex::HTML.instance_methods.include?(method_name) && instance_method(method_name).owner != Phlex::HTML
+					raise NameError, "👋 Redefining the method `#{name}##{method_name}` is not a good idea."
+				end
+
+				super
 			end
 		end
 
@@ -346,15 +358,6 @@ module Phlex
 			end
 
 			buffer
-		end
-
-		# This should be the last method defined
-		def self.method_added(method_name)
-			if method_name[0] == "_" && Phlex::HTML.instance_methods.include?(method_name) && instance_method(method_name).owner != Phlex::HTML
-				raise NameError, "👋 Redefining the method `#{name}##{method_name}` is not a good idea."
-			end
-
-			super
 		end
 	end
 end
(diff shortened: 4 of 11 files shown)


joelmoss:
You thought about providing `HTML` as both a module and class? Allowing us to choose the best tool for the job.

aaronjensen:
@joelmoss When would you use a class if a module was provided? I.e., what criteria would you use?

joelmoss:
I've actually created a few generic views that provide both a module and class. Usually I would create a new view and inherit from the provided class. But sometimes I may have another class I would prefer to inherit, for example to share some common functionality. In that case, I would include the module.

I've found this really useful and very versatile.

fixtures/view_helper.rb
```diff
@@ -10,7 +10,7 @@ def self.extended(parent)
 
 	def view(&block)
 		let :view do
-			Class.new(Phlex::HTML, &block)
+			Class.new { include Phlex::HTML }.tap { |c| c.class_eval(&block) }
```
aaronjensen, line 13: I *think* this is effectively the same thing: 

```rb
Class.new do
  include Phlex::HTML
  class_exec(&block)
end
```

aaronjensen:
To be honest, I'm not sure I follow. Modules are for sharing behavior. I would only extend a class if I had a concrete class that was in use and I needed to specialize it in some way. Using classes to share common functionality is usually indicative of a design problem.

If a module were provided, a person could include it into any class and that class could be extended if necessary, but modules can also be included in other, more specialized modules:

```rb
module SomeModule
  def self.included(cls)
    cls.class_exec do
      include Phlex::HTML
    end
  end

  def some_method
    # ...
  end
end
```

That said, it may provide more traceability and clarity to just do this:

```rb
class SomeClass
  include Phlex::HTML
  include SomeModule

  # ...
end
```

joeldrapper:
There are definitely drawbacks to this approach. You mentioned traceability — it would be impossible for a language server like Solargraph to follow.

joelmoss:
> Using classes to share common functionality is usually indicative of a design problem.

The Ruby world inherits from other classes all over the place, simply to use some common functionality. Think about `ApplicationController` or `ApplicationRecord` in Rails, and `ApplicationView` for Phlex.

All I'm saying is that there is a time and a place for modules and classes, so it would be nice to provide both, which is easy to do. But if not, it's easy enough to build it in userland too.

aaronjensen:
> The Ruby world inherits from other classes all over the place, simply to use some common functionality. Think about ApplicationController or ApplicationRecord in Rails, and ApplicationView for Phlex.

These are both anti-patterns. They stem from design mistakes in Rails. Can you instantiate a controller in Rails and test it? The design is flawed, but the author is anti-TDD and anti-testability and ultimately that leads to being anti-design. Unfortunately, those mistakes are now "norms", but it doesn't mean we need to propagate them if we can see them as such.

aaronjensen:
> it would be impossible for a language server like Solargraph to follow.

I can't speak to that as I don't use it. It may be that more sophisticated typings can cover it or it may be that solargraph isn't needed for something like phlex whose api is pretty much just html.

joelmoss:
> These are both anti-patterns.

I think we'll have to agree to disagree there. I use both modules and classes everywhere depending on the use case, and there are pros and cons to each.

The problem with anti-patterns is that they are highly opinionated 😏

joeldrapper:
I don't think an abstract class is an anti-pattern — and I have no problem testing it — that doesn't mean it's the right pattern for `Phlex::HTML`. Perhaps a module would be better here.

I don't think we can provide both without finding appropriate names. We could go with `Phlex::HTML::Base` for the class and `Phlex::HTML` for the module, but that's still a huge breaking change. I’m very cautious to introduce a change like that at this point.

joelmoss:
What was your original reasoning behind the change to a module?

joeldrapper:
More flexibility. You could include a module in another type of class. A simple example:

```ruby
MyView = Data.define(:foo, :bar) do
  include Phlex::HTML
end
```

joelmoss:
> We could go with `Phlex::HTML::Base` for the class and `Phlex::HTML` for the module...

FWIW, I would favour the above. Maybe for a v2 release though due to it being a major breaking change.

joeldrapper:
I think realistically this isn't something that can come before 2.0 if at all.

aaronjensen:
> I think we'll have to agree to disagree there. 

Works for me :)

> The problem with anti-patterns is that they are highly opinionated 😏

Well, this is truly an opinion. And I'm not trying to throw shade here but I'll just say that in my experience, this opinion is highly correlated with designs that cause real problems to productivity. 

Afference, efference, cohesion, and coupling are measurable and real and have real impact. But these things are subtle knowledge and if it's not understood I can't really argue it.

aaronjensen:
> I don't think an abstract class is an anti-pattern

It's not in a language that doesn't have modules (and those often actually have a material notion of an abstract base class). The problem in Ruby is that it has modules for sharing behavior, and I'm yet to encounter a reason for preferring an abstract base class that is more than some form of fixation (like a Java developer who is used to them wanting them in Ruby and not being familiar w/ how to use modules in the same way). If anyone has a use case though, I'd be very interested in it, so that I can expand my own understanding.

joeldrapper:
This was worth exploring but it's not happening in 1.x. We should seriously reconsider it for 2.0 one day.

zavan:
Just leaving a comment to say that I just felt the need to use it as a module:

```ruby
class BaseRenderer
  # Some behaviour shared by all renderers.
end

class HtmlRenderer < BaseRenderer
  include Phlex::HTML

  # Do HTML stuff with Phlex
end

class JsonRenderer < BaseRenderer
  # Do JSON stuff
end

# Other classes that extend Html and Json renderers
```

I could move the BaseRederer stuff to a module and include it in Html and Json renderers instead, but I think the above would make more sense.

aaronjensen:
Just curious, what is it about BaseRender that makes it more suitable to be a class rather than a module?

zavan:
@aaronjensen in my real code it has class methods that are inherited by the subclasses. The class methods could go into a module, but that requires extra code (def self.included...). It being a class also goes well with Rails' pattern of Base classes.

aaronjensen:
Copy that, thanks.

joeldrapper:
@zavan thanks for the use-case. My concern here is the module interface would be tied to whatever number of "renderable" interfaces might be including it. Whatever you include for JSON renderers might not be compatible with the way the HTML renderer works, and the way that works could change in the future.

Phlex now supports `HTML` and `SVG` with a common `SGML` class. It's likely we'll go further and have a `Renderable` class or module that `SGML` inherits from, allowing us to define `Phlex::JSON`, etc.

PR #493, opened 2023-02-16
Phlex::SVG

davekaro:
Following up on https://github.com/joeldrapper/phlex/pull/492. Mostly adding the remaining SVG tags. Still needs more tests and work.


joeldrapper:
This is looking really great. I will follow up with a closer review soon.

lib/phlex/xml.rb
```diff
@@ -0,0 +1,315 @@
+# frozen_string_literal: true
+
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("3.0")
+	using Phlex::Overrides::Symbol::Name
+end
+
+module Phlex
+	class XML
```
joeldrapper, line 8: Can we rename this `SGML`? https://en.wikipedia.org/wiki/Standard_Generalized_Markup_Language

joeldrapper:
I think this could do with a fair amount of refactoring, but it's a huge PR and it works and I’d like to merge it sooner rather than later so it doesn't get out of sync with other work.

lib/phlex/xml.rb
marcoroth: I feel like `SGML` is weird, because no one knows this by heart. Also Nokogiri uses `XML` throughout their lib for representing XML and HTML nodes

davekaro: Users of Phlex probably don't need to know about the `SGML` class, right? I think `SGML` makes sense, and could see a future `Phlex::XML` to generate strict XML.

joeldrapper: Yes, I would like to have a `Phlex::XML` class for working with strict XML with support for namespaces.

marcoroth: Fair enough 👍🏼

PR #507, opened 2023-02-27
Shim `Phlex::HTML#text` method including deprecation warning

marcoroth:
That way it won't break existing HTML views in applications which are currently relying on `text`

### DIFF
--- a/test/phlex/view/text.rb
+++ b/test/phlex/view/text.rb
@@ -38,4 +38,17 @@ describe Phlex::HTML do
 			expect(output).to be == "2.0"
 		end
 	end
+
+	with "deprecated text method" do
+		view do
+			def template
+				text "Depreacted"
+			end
+		end
+
+		it "renders content and prints deprecation warning" do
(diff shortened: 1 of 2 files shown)


test/phlex/view/text.rb
```diff
@@ -38,4 +38,17 @@ def template
...
+
+	with "deprecated text method" do
+		view do
+			def template
+				text "Depreacted"
+			end
+		end
+
+		it "renders content and prints deprecation warning" do
+			expect(example).to receive(:warn).with("DEPRECATED: The `text` method has been deprecated in favour of `plain`. Please use `plain` instead. The `text` method will be removed in a future version of Phlex. Called from: test/phlex/view/text.rb:47:in `template'")
```
marcoroth, line 50: Seems like it's super hard to check if something gets printed to STDERR with Sus. At least I haven't figured out how to make it pass in CI. Also the matchers doesn't show a diff why the assertion failed 🤔  This test passes locally though.

If you have any idea, let me know

PR #512, opened 2023-03-03
Change strategy of determining the origin of an element method

willcosgrove:
Previously the method would be checked to see if it's owner was Phlex::HTML, but this doesn't automatically work across multiple Phlex::SGMLs. Now we have an obnoxiously named alias to cue off of to determine if a method is a real Phlex element.


lib/phlex/sgml.rb
```diff
@@ -304,5 +304,14 @@ def capture(&block)
 
 			buffer
 		end
+
+		# This should be the last method defined
+		def self.method_added(method_name)
```
willcosgrove, line 309: @joeldrapper I didn't know if we wanted to move this guardrail to `SGML`, or duplicate it across `HTML` and `SVG`. Due to the strategy change, it is inheritable. But inheriting it also means that every method defined in `HTML` and `SVG` are needlessly incurring the penalty of checking this guardrail. But maybe the one time performance cost is worth the extra dryness.

lib/phlex/elements.rb
```diff
@@ -32,6 +32,8 @@ def #{element}(**attributes, &block)
 			end
 
 			alias_method :_#{element}, :#{element}
+			alias_method :__phlex_#{element}__, :#{element}
```
willcosgrove, line 35: One very minor performance thought I had here: if we changed the element method definition around so that `__phlex_#{element}__` was the original method, and `#{element}` was the alias, then we could check if a method is the real Phlex element method by checking `UnboundedMethod#original_name` instead of comparing two `UnboundedMethod` instances. That would reduce the amount of UnboundedMethod allocations when checking to see if something can be optimized.

joeldrapper: This would be even better 👍

lib/phlex/sgml.rb
joeldrapper: Shall we put it in a module and have HTML and SVG extend that module as the last thing?

PR #515, opened 2023-03-05
Pre-process attributes with a class-isolated attribute cache

### LINKED ISSUE #467: Provide an official attributes hook
A few people have expressed an interest in overriding `__attributes__` in order to prepare or validate their tag attributes in some unique way. This is not a supported use-case and was recently broken by #459.

I’m opening this issue to explore the possibility of exposing an official hook like our hooks for `before_template`, `after_template` and `around_template`.

### Performance

Processing a Hash of attributes is by far the slowest thing in Phlex. To manage this, we store a cache of the processed attributes on the first render of any view class. We can then take advantage of this cache in subsequent renders if the attributes are consistent, which is usually the case.

Any hook should ideally happen after the cache check, with any customisations of the attributes included in the cache.

Additionally, adding this hook to Phlex should not slow down normal renders that don't implement the hook.

### Safety

Phlex uses the return value of `__attributes__` in the HTML output, which means if you override `__attributes__` and return a `String` rather than calling and returning `super`, you bypass important HTML-safety checks. Any official attributes hook must not leave the application vulnerable if `super` is not called and returned by the override.

joeldrapper:
Allow users to define custom attribute processing logic on component classes by defining `process_attributes(**attributes)`.

When this method is defined, the attributes cache will include the component class as part of the cache key, so the cache is isolated for each distinct class. While attribute processing can be different for each component class, it should _always_ be deterministic because the cache will still be based on the original unprocessed Hash.

### Performance

Introducing this feature brings ~1% performance cost for classes that don't opt in to using the feature by defining `process_attributes` and ~10% performance cost for classes that do use the feature.

### Security

The contract with `process_attributes` is that it takes keyword arguments and returns a Hash. Since this Hash then goes through Phlex own `__attributes__` processor, it doesn't bypass any built in security features.

Closes #467

### DIFF
--- a/lib/phlex/elements.rb
+++ b/lib/phlex/elements.rb
@@ -20,11 +20,11 @@ module Phlex::Elements
 			def __phlex_#{element}__(**attributes, &block)
 				if attributes.length > 0
 					if block_given?
-						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[process_attributes? ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << ">"
+						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[respond_to?(:process_attributes) ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << ">"
 						yield_content(&block)
 						@_target << "</#{tag}>"
 					else
-						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[process_attributes? ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << "></#{tag}>"
+						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[respond_to?(:process_attributes) ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << "></#{tag}>"
 					end
 				else
 					if block_given?
@@ -53,7 +53,7 @@ module Phlex::Elements
 
 			def __phlex_#{element}__(**attributes)
 				if attributes.length > 0
-					@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[process_attributes? ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << ">"
+					@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[respond_to?(:process_attributes) ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << ">"
 				else
 					@_target << "<#{tag}>"
 				end
--- a/lib/phlex/elements.rb
+++ b/lib/phlex/elements.rb
@@ -20,11 +20,11 @@ module Phlex::Elements
 			def __phlex_#{element}__(**attributes, &block)
 				if attributes.length > 0
 					if block_given?
-						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[attributes.hash] || __attributes__(**attributes)) << ">"
+						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[process_attributes? ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << ">"
 						yield_content(&block)
 						@_target << "</#{tag}>"
 					else
-						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[attributes.hash] || __attributes__(**attributes)) << "></#{tag}>"
+						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[process_attributes? ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << "></#{tag}>"
 					end
 				else
 					if block_given?
@@ -53,7 +53,7 @@ module Phlex::Elements
 
 			def __phlex_#{element}__(**attributes)
 				if attributes.length > 0
-					@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[attributes.hash] || __attributes__(**attributes)) << ">"
+					@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[process_attributes? ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << ">"
 				else
 					@_target << "<#{tag}>"
 				end
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -72,10 +72,6 @@ module Phlex
 			buffer ? (buffer << target) : target
 		end
 
-		def process_attributes?
-			false
-		end
-
 		# Render another view
 		# @param renderable [Phlex::SGML]
 		# @return [nil]
@@ -250,13 +246,13 @@ module Phlex
 		# @api private
 		private def __attributes__(**attributes)
 			__final_attributes__(**attributes).tap do |buffer|
-				Phlex::ATTRIBUTE_CACHE[process_attributes? ? (attributes.hash + self.class.hash) : attributes.hash] = buffer.freeze
+				Phlex::ATTRIBUTE_CACHE[respond_to?(:process_attributes) ? (attributes.hash + self.class.hash) : attributes.hash] = buffer.freeze
 			end
 		end
 
 		# @api private
 		private def __final_attributes__(**attributes)
-			if process_attributes?
+			if respond_to?(:process_attributes)
 				attributes = process_attributes(**attributes)
 			end
 
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -72,6 +72,10 @@ module Phlex
 			buffer ? (buffer << target) : target
 		end
 
+		def process_attributes?
+			false
+		end
+
 		# Render another view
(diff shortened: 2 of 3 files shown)


joeldrapper:
@joelmoss if you want to try this out, you can define `process_attributes?` to return `true` and `process_attributes(**attributes)` to return a Hash of processed attributes. The cache should be isolated to the class.

lib/phlex/sgml.rb
```diff
@@ -72,6 +72,10 @@ def __final_call__(buffer = nil, target: +"", view_context: nil, parent: nil, &b
 			buffer ? (buffer << target) : target
 		end
 
+		def process_attributes?
```
joeldrapper, line 75: this should be private

lib/phlex/elements.rb
```diff
@@ -20,11 +20,11 @@ def register_element(element, tag: element.name.tr("_", "-"))
 			def __phlex_#{element}__(**attributes, &block)
 				if attributes.length > 0
 					if block_given?
-						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[attributes.hash] || __attributes__(**attributes)) << ">"
+						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[process_attributes? ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << ">"
```
joeldrapper, line 23: It's significantly cheaper to do this cache check here rather than always calling `__attributes__` and having that method do the cache check.

joeldrapper:
One thing I don't like about this solution is it binds pre-processing attributes with a class-isolated attribute cache, and you may want to pre-process attributes without a class-isolated attribute cache if your pre-processing is global for all view classes.

lib/phlex/sgml.rb
joelmoss: Is there a need for the additional `process_attributes?` method, as opposed to simply using `respond_to?(:process_attributes)`?

joeldrapper: `respond_to?` used to be much slower than calling a method that just returns `true` or `false`. However, I just benchmarked it again and it looks like YJIT makes it faster.

lib/phlex/elements.rb
```diff
@@ -20,11 +20,11 @@ def register_element(element, tag: element.name.tr("_", "-"))
 			def __phlex_#{element}__(**attributes, &block)
 				if attributes.length > 0
 					if block_given?
-						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[attributes.hash] || __attributes__(**attributes)) << ">"
+						@_target << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[respond_to?(:process_attributes) ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << ">"
```
willcosgrove, line 23: I wonder how brutal the performance hit would be to use `unbounded_method(:process_attributes).hash` instead of the class's hash.

I think that would make it so the attribute cache is only busted if you redefine `process_attributes`.

I'll try and benchmark it later today if I get time.

willcosgrove: It's 3 times slower 😢

```
ruby 3.2.1 (2023-02-08 revision 31819e82c8) +YJIT [arm64-darwin22]
self.class.instance_method.hash
                       682.322k i/100ms
     self.class.hash     1.605M i/100ms
self.class.instance_method.hash
                          9.102M (± 2.2%) i/s -     45.716M in   5.025309s
     self.class.hash     28.106M (± 1.9%) i/s -    141.246M in   5.027368s

Comparison:
     self.class.hash: 28106052.4 i/s
self.class.instance_method.hash:  9101860.8 i/s - 3.09x  (± 0.00) slower
```

I think basing the cache key off the method implementation is the more correct caching behavior, but since the cache is just there to be fast, I'm not sure that having ideal caching behavior at the expense of speed is a good trade off.

I don't intend to add a `process_attributes` method on my `ApplicationView`, so I'd be fine to see it ship like this. It would be worthwhile to point out in the documentation.

willcosgrove: Hmm, it doesn't appear to affect performance much though in the bench.rb file. I modified the Layout fixture to define `process_attributes` like this:

```ruby
def process_attributes(a)
  a.transform_keys(&:upcase)
end
```

And it's running at basically the same speed:

## Process Attributes defined on Layout + instance_method.hash
```
env RUBY_YJIT_ENABLE=1 ruby bench.rb
ruby 3.2.1 (2023-02-08 revision 31819e82c8) +YJIT [arm64-darwin22]
                Page     9.990k i/100ms
                Page    105.026k (± 2.3%) i/s -    529.470k in   5.044352s
```

## Process Attributes defined on Layout + class.hash
```
ruby 3.2.1 (2023-02-08 revision 31819e82c8) +YJIT [arm64-darwin22]
                Page    10.719k i/100ms
                Page    104.551k (± 3.5%) i/s -    525.231k in   5.030099s
```

## Process Attributes _not_ defined, with instance_method.hash
```
ruby 3.2.1 (2023-02-08 revision 31819e82c8) +YJIT [arm64-darwin22]
                Page    10.265k i/100ms
                Page    104.474k (± 2.4%) i/s -    523.515k in   5.014144s
```

## Process Attributes _not_ defined, with class.hash
```
ruby 3.2.1 (2023-02-08 revision 31819e82c8) +YJIT [arm64-darwin22]
                Page    10.338k i/100ms
                Page    103.636k (± 2.3%) i/s -    527.238k in   5.090344s
```

I would have expected a 3x difference in the microbenchmark to be more measurable in the macrobenchmark. But maybe it's not a huge penalty 🤷? Or maybe I tested it wrong 😬

willcosgrove: _Narrator: He tested it wrong._

I didn't notice that the `bench.rb` is requiring the phlex gem, and not the local relative phlex project. So none of my changes were being benchmarked 🙃 

It's about half as fast to check the method's hash than to check the class's hash. So I think class check is the way to go.

joeldrapper: You can run `bundle exec ./bench.rb` to use the bundled gems.

lib/phlex/sgml.rb
joelmoss: Gotta love YJIT! 😁

joeldrapper: I still wonder if changing the caching strategy should be a separate thing in case someone wants to use `process_attributes` without using class-isolated attribute caching since that part costs ~10% and isn't necessary if you have one global, deterministic processing strategy.

willcosgrove: That's an interesting idea: a way to opt-in to an isolated attribute cache. Similar to how you can opt in to deferred rendering.

joelmoss:
ok, so this all works perfectly as far as I'm concerned 👌

joeldrapper:
> One thing I don't like about this solution is it binds pre-processing attributes with a class-isolated attribute cache, and you may want to pre-process attributes without a class-isolated attribute cache if your pre-processing is global for all view classes.

I’ve been thinking more about this and I do think it makes sense to switch to class-isolated attribute cache keys when `process_attributes` is defined. However, it may be appropriate to provide a configuration setting for a global attribute processor that wouldn't affect the cache keys.

I could see something like this in the future:

```ruby
Phlex.configure do |c|
  c.process_attributes do |attributes|
    ...
  end
end
```

joelmoss:
ok, so I spoke too soon :( The shared attribute cache is still biting me.

The shared cache is only a problem for me, because I am also using `process_attributes` to call code that should be called every time a view is rendered, but only if specific attribute conditions are met.

But because `process_attributes` is only called the once, no matter how many instances of the view are created and rendered, my code is only called the once.

Yes there are the `[before|after]_template` hooks where I could call my code. But that won't work because I still only want to call my code if the my attributes conditions are met.

So this new hook is great for processing attributes, but not for triggering actions based on attributes used, as the hook only gets called the once no matter how many view instances are rendered.

joeldrapper:
I can't think of a solution to this without losing all the benefits of caching which is a ~52% performance cost. Can you help me understand what you need this for? There might be another way.

joelmoss:
so I think I found a solution that is working... so far 😁

joelmoss:
I'm still using `process_attributes`. I have to, because that is the only way to know about the attributes.

When attribute conditions are met, I append to a class attribute (defined as a `Set` using `attr_accessor` on the view class, not instance). I then call my code from the `after_template` hook when the class attribute is populated.

Something like this...

```ruby
class_methods do
  attr_accessor :side_load_cache
end

def before_template
  self.class.side_load_cache ||= Set.new
  super
end

def process_attributes(**attributes)
  if attributes.key?(:class) && attributes[:class].starts_with?('@')
    self.class.side_load_cache << "path/to/#{attributes[:class]}"
  end

  attributes
end

def after_template
  super

  self.class.side_load_cache&.each do |path|
    Proscenium::SideLoad.append! path
  end
end
```

So far it works, and I avoid duplicates by using `Set`.

How would a global hook be better?

joeldrapper:
Oh that's great, that's a nice solution. A global hook is just simpler, faster and easier to maintain since it can use the global attributes cache. If there are use-cases for `process_attributes`, I’m happy to keep it.

You may want to use `Concurrent::Set` from the `concurrent-ruby` project if you want this to be thread-safe. `concurrent-ruby` is already a dependency of Phlex, so you may as well.

joelmoss:
Global hook sounds good, but we would lose its context, such as the view instance.

joeldrapper:
If the view instance is helpful to someone, I’m happy to keep `process_attributes` as it is.

PR #531, opened 2023-03-10
Improved element register

joeldrapper:
### Listing registered elements
`Phlex::Element` classes (such as `Phlex::HTML::StandardElements`) now respond to `registered_elements` with a `Concurrent::Map(method_name => Symbol, tag => String)` of all their registered elements.

### Checking if a method is a registered element
`Phlex::SGML` classes (`Phlex::HTML` and `Phlex::SGML`) now also respond to `registered_element?(method_name => Symbol) => bool`. This works by looking up the method and asking its _owner_ if it's a registered element.

Note: this second method is part of Phlex’ private API.

### DIFF
--- a/lib/phlex/element_clobbering_guard.rb
+++ b/lib/phlex/element_clobbering_guard.rb
@@ -3,11 +3,11 @@
 module Phlex
 	module ElementClobberingGuard
 		def method_added(method_name)
-			if method_name[0] == "_" && private_instance_methods.include?(:"__phlex#{method_name}__")
+			if method_name[0] == "_" && element_method?(method_name[1..].to_sym)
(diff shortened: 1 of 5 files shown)


joeldrapper:
@marcoroth the `registered_elements` method should be a lot better than `slow_registered_elements` as it includes the tag name as well as the method name.

marcoroth:
Nice, thank you!

marcoroth:
I just updated `phlexing` to use `1.6.1` in https://github.com/marcoroth/phlexing/commit/8ff882e7f3e8951f42e354852a01e8866d031070 in order to make use of this method. Works flawlessly, thank you!

Issue #518, opened 2023-03-06
Changes to `text`/`plain` method resulting in content output issues


Bumppoman:
This is the same issue occurred in #393.  Previously, the default behavior of `text(object)` would fall back to `object.to_s` (i.e. the `text` would force text output without an explicit `to_s`.). This has changed again in 1.5/1.5.1.

I would again argue that the previous behavior was desired (if you call this method, you are declaring that you want some output if the output is available).  I think an explicit `to_s` is redundant.

joeldrapper:
We removed the implicit `to_s` because a lot of objects respond to `to_s` even when they shouldn't really be expected to output anything. Instances of `Enumerator` return something like this `"#<Enumerator:0x0000000102f5f2c8>"` and an instance of an empty `Array` returns `[]`. That could catch you out if you have a block that iterates over an array, but sometimes it's empty.

```ruby
ul do
  list.each do |item|
    li { item }
  end
end
```

Because the block passed to `ul` doesn't add any `li` tags when empty, `ul` will look at the return value instead. If we call `to_s` on the return value, we'll get `[]`, so the output will be `<ul>[]</ul>`.

If you have specific types of object that you want to format for output, you can define the `format_object` method. When Phlex doesn't know how to format an object, it will pass it to this method expecting it to return a string.

```ruby
def format_object(object)
  case object
  when Money
    object.format
  when Date, DateTime
    object.strftime("%Y-%m-%d")
  else
    super
  end
end
```

You could also return `object.to_s` in this `else` condition if you want that to be the default; otherwise, call `super` from `else` to fall back to Phlex’ default.

Bumppoman:
I agree with the foregoing as to why the element methods don't have an implicit `to_s` but what else would you expect the output of `plain(object)` to be?  It seems that the new behavior (returning either `""` or `nil`) would be less obvious than defaulting to the string value.

joeldrapper:
That’s a good point `plain` could behave differently. If `format_object` returns `nil` for the object, we could try calling `to_s` instead. 🤔

PR #545, opened 2023-03-27
Raise when trying to output an unknown object using `plain`

### LINKED ISSUE #544: Raise an error when `plain` is called with an object that can’t be formatted
We should extract a specialised private `__text__` method for `yield_content` that doesn’t raise.

willcosgrove:
The previous behavior was to output nothing, which might be unexpected, and could potentially go unnoticed. Now an `ArgumentError` will be raised along with an explanation of how to fix it:  which is to define your own `format_object` definition that knows how to convert the object to a string.

An implicit conversion to a string via `to_s` is being considered, but given that many objects in ruby implement `to_s` in a way that is not intended to be user facing, this could cause issues. Since the current behavior (of silently doing nothing) definitely seems wrong, we are starting with this as a first step. With a possible next step of coercing to a string via `to_s`.

I also moved the deprecated `text` test to the top of the file, because it does a check that involves the line number. I figured it might make the test more stable to move it to the top.

Closes #544, #518

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -127,7 +127,7 @@ module Phlex
 		end
 
 		# Output text content. The text will be HTML-escaped.
-		# @param content [String, Symbol, Integer, void] the content to be output on the buffer. Strings, Symbols, and Integers are handled by <code>plain</code> directly, but any object can be handled by overriding <code>format_object</code>
+		# @param content [String, Symbol, Integer, void] the content to be output on the buffer. Strings, Symbols, and Integers are handled by `plain` directly, but any object can be handled by overriding `format_object`
 		# @return [nil]
 		# @see #format_object
 		def plain(content)
@@ -241,7 +241,7 @@ module Phlex
 			nil
 		end
 
-		# Determines if the component should render. By default, it returns <code>true</code>.
+		# Determines if the component should render. By default, it returns `true`.
 		# @abstract Override to define your own predicate to prevent rendering.
 		# @return [Boolean]
 		def render?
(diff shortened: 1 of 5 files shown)


lib/phlex/sgml.rb
```diff
@@ -127,7 +127,9 @@ def __final_call__(buffer = +"", context: Phlex::Context.new, view_context: nil,
 		end
 
 		# Output text content. The text will be HTML-escaped.
+		# @param content [String, Symbol, Integer, void] the content to be output on the buffer. Strings, Symbols, and Integers are handled by <code>plain</code> directly, but any object can be handled by overriding <code>format_object</code>
```
joeldrapper, line 130: We can use markdown in these doc strings now that's configured in yardopts so you can use <code>\`</code> fences instead of `<code>`.

test/phlex/view/text.rb
```diff
@@ -3,6 +3,19 @@
...
+				text "Depreacted"
+			end
+		end
+
+		it "renders content and prints deprecation warning" do
+			expect(example).to receive(:warn).with("DEPRECATED: The `text` method has been deprecated in favour of `plain`. Please use `plain` instead. The `text` method will be removed in a future version of Phlex. Called from: test/phlex/view/text.rb:9:in `template'")
+			expect(output).to be == "Depreacted"
+		end
+	end
+
```
joeldrapper, line 18: We can remove this test and the `text` method in this PR if you want. It needs to go in the next minor release anyway.

joeldrapper (approved):

PR #546, opened 2023-03-28
Support for Array and Set attribute values

willcosgrove:
This adds support for Array or Set attribute values like this:

```ruby
div(class: ["bg-red-500", "rounded"])
# or
div(class: Set["bg-red-500", "rounded"])
```

Array or Set attribute values will be treated like a token list, and its members will be joined with a space. This works as long as all elements are either strings or symbols.

Additionally, this change removes the implicit conversion of attribute values to a string by calling `to_s` on anything that is not recognized. Now attribute values that are not either a String, a Symbol, an Array, a Set, or a Hash with values of one of these types, will raise an ArgumentError.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -397,12 +397,6 @@ module Phlex
 					buffer << " " << name << '="' << ERB::Escape.html_escape(v) << '"'
 				when Symbol
 					buffer << " " << name << '="' << ERB::Escape.html_escape(v.name) << '"'
-				when Array
-					if v.all? { |el| el.is_a?(String) || el.is_a?(Symbol) }
-						buffer << " " << name << '="' << ERB::Escape.html_escape(v.join(" ")) << '"'
-					else
-						raise ArgumentError, "Element attributes must be either a Boolean, a String, a Symbol, an Array of Strings or Symbols, or a Hash with values of one of these types"
-					end
 				when Hash
 					__build_attributes__(
 						v.transform_keys { |subkey|
@@ -412,6 +406,12 @@ module Phlex
 							end
 						}, buffer: buffer
 					)
+				when Array
+					if v.all? { |el| el.is_a?(String) || el.is_a?(Symbol) }
+						buffer << " " << name << '="' << ERB::Escape.html_escape(v.join(" ")) << '"'
+					else
+						raise ArgumentError, "Element attributes must be either a Boolean, a String, a Symbol, an Array of Strings or Symbols, or a Hash with values of one of these types"
+					end
 				else
 					raise ArgumentError, "Element attributes must be either a Boolean, a String, a Symbol, an Array of Strings or Symbols, or a Hash with values of one of these types"
 				end
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -391,15 +391,19 @@ module Phlex
 				end
 
 				case v
-				in true
+				when true
 					buffer << " " << name
-				in String
+				when String
 					buffer << " " << name << '="' << ERB::Escape.html_escape(v) << '"'
-				in Symbol
+				when Symbol
 					buffer << " " << name << '="' << ERB::Escape.html_escape(v.name) << '"'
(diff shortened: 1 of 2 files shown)


lib/phlex/sgml.rb
```diff
@@ -406,8 +406,14 @@ def __build_attributes__(attributes, buffer:)
 							end
 						}, buffer: buffer
 					)
+				when Array
+					if v.all? { |el| el.is_a?(String) || el.is_a?(Symbol) }
+						buffer << " " << name << '="' << ERB::Escape.html_escape(v.join(" ")) << '"'
+					else
+						raise ArgumentError, "Element attributes must be either a Boolean, a String, a Symbol, an Array of Strings or Symbols, or a Hash with values of one of these types"
+					end
```
joeldrapper, line 414: I think we could probably skip the type check here and just `join`.

```suggestion
					buffer << " " << name << '="' << ERB::Escape.html_escape(v.join(" ")) << '"'
```

lib/phlex/sgml.rb
```diff
@@ -406,8 +406,10 @@ def __build_attributes__(attributes, buffer:)
 							end
 						}, buffer: buffer
 					)
+				when Array
```
willcosgrove, line 409:
```suggestion
				when Array, Set
```

@joeldrapper can you think of any negatives to also supporting `Set` here?

willcosgrove: Looks like `Set#join` was added in Ruby 3.0 😕

Do you want to revert this, or just let it go untested and note that Sets are only allowed on Ruby >=3?

hangsu:
This is something we've implemented at Clearscope by overriding `__final_attributes__`, and have found it to work well for conditional tokens.

My suggestion would be to `compact` before `join(" ")` to avoid double spaces in the case of: 
```ruby
class: ["foo", ("bar" if false), "baz"]
```

joeldrapper:
I don't know if compacting the list is worth the cost because we'd need to iterate through the list two times.

You can always use a Set instead of an Array, which will be compacted on the fly. I haven't benchmarked this so I don't know what the real impact is, but I expect it's not worth it. A few extra spaces in the HTML output should be fine.

willcosgrove:
I don't have an explanation for this, but when I benchmarked `array.join(' ')` against `array.compact.join(' '), the `compact.join` is consistently >2x faster.

```ruby
array = ["bg-brand-700", "text-white", nil, "hover:text-white", "border", "border-brand-700", "hover:border-brand-800", nil, "hover:bg-brand-800", "disabled:bg-brand-700/70", nil, nil, "disabled:border-transparent", "disabled:hover:bg-brand-700/70", "disabled:hover:border-transparent", "disabled:cursor-not-allowed", nil]

Benchmark.ips do |x|
  x.report("array.join(' ')") { array.join(' ') }
  x.report("array.compact.join(' ')") { array.compact.join(' ') }
  x.compare!
end
```

joeldrapper (approved):

hms:
@joeldrapper is there a reason for "removes the implicit conversion of attribute values to a string by calling to_s"?  It seems like it should handle integer values as that's a very common use case.  For example, passing a  value to a stimulus Numeric value attribute.  

While it's trivial to make this change in my code base, I not sure I understand why so many other typical use cases are handled, and a couple non-typical use cases (that are ergonomically very nice), are handled and the integer use case is pushed back to the developer.

willcosgrove:
I think handling `Numeric` here is a good idea. The intention with this change was to not do not-obvious things, like convert objects to strings. Previously arrays would get converted to a string which was probably not what would be intended (`<div class="[\"text-lg\", \"bg-white\"]"></div>`).

But I think there's only one thing you'd reasonably expect to have happen for a number passed as an attribute.

joeldrapper:
Let's add support for integers and floats like we do with text.

hms:
@joeldrapper @willcosgrove 

A quick shout-out for such a fast response.  Thank you again, and again, and again for all of the help, responsiveness to my questions / requests, and a great product.

Hal

PR #558, opened 2023-04-18
Support 0-argument lambdas

joeldrapper:
`SGML#render` now check the arity of the given Proc in order to support zero-argument lambdas. This make it compatible with “stabby lambda” literals `-> {}`.


willcosgrove (approved):
I love my stabby lambdas ->❤️

PR #561, opened 2023-04-19
Support numeric attribute values

### LINKED ISSUE #560: Support integer and float attribute values

joeldrapper:
This PR adds support for numeric attribute values (Integer and Float) as well as introducing a fallback to call `to_str` on other objects.

Closes #560

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -337,7 +337,7 @@ module Phlex
 			when Symbol
 				@_context.target << ERB::Escape.html_escape(content.name)
 			when Integer
-				@_context.target << ERB::Escape.html_escape(content.to_s)
+				@_context.target << content.to_s
 			when nil
 				nil
 			else


willcosgrove:
Sorry, I thought I was clicking a "review" button 😇 

This looks good to me! I missed it and made my own branch just now. But yours looks better, I missed that we can remove the HTML escaping for this branch. I do have a test if you want to copy it into yours. https://github.com/phlex-ruby/phlex/compare/integer-float-attributes

joeldrapper:
Sorry, I should have assigned the issue to myself. I'll copy your test.

joeldrapper:
@willcosgrove what are your thoughts on falling back to `to_str` here? This seems to be more suitable than `to_s`. See https://blog.appsignal.com/2018/09/25/explicitly-casting-vs-implicitly-coercing-types-in-ruby.html

willcosgrove (approved):
`to_str` looks appropriate here to me. Let's go with that! Good find

PR #566, opened 2023-04-21
`flush` during `capture`

willcosgrove:
If you issue a `flush` while inside of a `capture` block, the flush will happen to the calling component's buffer and not to the string being captured.

This fixes the issue by not only changing out the `@context.target` but also the buffer for the duration of the `capture` block.

I encountered this issue because I was capturing the output of a full HTML document, which was being generated by another Phlex view), for the purposes of passing it as a `srcdoc` to an `<iframe>`. But the closing `</head>` tag in the iframe document was pushing the captured buffer up to that point onto the parent document.

I benchmarked this change and the difference falls within error.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -182,7 +182,20 @@ module Phlex
 		def capture(&block)
 			return "" unless block
 
-			@_context.with_target(+"") { yield_content(&block) }
+			new_buffer = +""
+
+			begin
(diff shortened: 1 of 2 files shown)


willcosgrove:
I'm going to close this in favor of the approach in #567. As I mentioned in that PR, this approach does not solve the case of a `render` and a `capture` happening in different contexts.

PR #567, opened 2023-04-21
Fix `flush` during `capture` by storing capturing state in `Context`

willcosgrove:
This is a 2nd attempt to solve the issue of flushing inside a capture. My attempt in #566 partially solves the issue, but the problem persists if the flush happens inside of a component that is being rendered by a component that is being captured by another component. So imagine a setup like so:

```rb
class Page < Phlex::HTML
  def template
    render Previewer do
      render FlushingComponent
    end
  end
end

class Previewer < Phlex::HTML
  def template
    srcdoc = capture { yield } if block_given?
    
    iframe srcdoc:
  end
end

class FlushingComponent < Phlex::HTML
  def template
    h1 { "first" }
    flush
    h1 { "second" }
  end
end
```

`Page` is rendering `FlushingComponent`, and `Previewer` is trying to capture it.

This PR attempts to fix the problem the same way this was fixed for `capture` in #526. When a view is called, we take the provided buffer and wrap it in a `Phlex::Context`. This allows all nested components to share the same wrapped buffer, which allows us to take advantage of `with_target` to switch out the underlying buffer during a capture.

## Potential for renaming

I think if this general strategy is acceptable, we may want to brainstorm new names for either `Phlex::Context` or for the `context` buffer shared between components during rendering, or for both.

A possible other name for `Phlex::Context` might be `Phlex::Box` as it reminds me of a [Rust Box](https://doc.rust-lang.org/std/boxed/index.html). `Phlex::BufferWrapper` or `Phlex::Container` I think also work.

If we were to rename the `context` ivar, I might suggest `internal_buffer`. We can thing of `buffer` as something that points outside of Phlex, like a Rack stream. It should be treated as potentially costly to write to. Whereas `internal_buffer` is where we write to most of the time, as we can be sure of its performance cost.

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -4,18 +4,22 @@
 class Phlex::Context
 	def initialize
 		@target = +""
+		@capturing = false
 	end
 
-	attr_accessor :target
+	attr_accessor :target, :capturing
 
-	def with_target(new_target)
+	def capturing_into(new_target)
 		original_target = @target
+		original_capturing = @capturing
 
 		begin
 			@target = new_target
+			@capturing = true
 			yield
 		ensure
 			@target = original_target
+			@capturing = original_capturing
 		end
 
 		new_target
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -182,13 +182,15 @@ module Phlex
 		def capture(&block)
 			return "" unless block
 
-			@_context.with_target(+"") { yield_content(&block) }
+			@_context.capturing_into(+"") { yield_content(&block) }
 		end
 
 		private
 
 		# @api private
 		def flush
+			return if @_context.capturing
+
 			target = @_context.target
 			@_buffer << target.dup
 			target.clear
@@ -240,7 +242,7 @@ module Phlex
 		def __vanish__(*args)
 			return unless block_given?
 
-			@_context.with_target(BlackHole) { yield(*args) }
+			@_context.capturing_into(BlackHole) { yield(*args) }
 
 			nil
 		end
--- a/test/phlex/view/capture.rb
+++ b/test/phlex/view/capture.rb
@@ -38,4 +38,67 @@ describe Phlex::HTML do
 			expect(example.captured).to be == "<h1>Hello</h1>"
 		end
 	end
+
+	with "a call to flush inside the block" do
+		view do
+			attr_accessor :captured
+
+			def template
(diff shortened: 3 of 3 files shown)


joeldrapper:
Another approach that might work would be to have the context know if we're capturing or not. Something like this:

```ruby
class Phlex::SGML
  def capture
    context.capturing do
      # current implementation of capture
    end
  end

  def flush
    return if context.capturing
    # current implementation of flush
  end
end

class Phlex::Context
  attr_reader :capturing

  def capturing
    original_capturing = @capturing
    @capturing = true
    yield
    @capturing = original_capturing
  end
end
```

willcosgrove:
I like that, since `flush` doesn't really have a use when capturing.

What would you think about combining `Context#capturing` with `Context#with_target`? I don't think we currently would use one without the other. It looks like we use `with_target(BlackHole)` when vanishing, maybe we could make the combination of the two `Context#capturing_into(target)`? 

This doesn't need to all happen in one PR, but any thoughts on renaming "context"? I kind of like `buffer -> output_buffer` and `context -> internal_buffer`, with the distinction for the two buffers being explained as: the `output_buffer` can be set externally, and is treated as potentially costly to write to, whereas the `internal_buffer` is managed internally and has a fixed known write cost. `Phlex::Context` can become `Phlex::Buffer`, and `internal_buffer` would be an instance of that. `output_buffer` would remain as is: passed in from the caller but defaulting to a string.

joeldrapper:
I think we could rename `target` to `inernal_buffer` and `buffer` to `output_buffer` but I don't think we should rename context.

The context object represents a discrete render context shared between each component of a full render. At the moment, it's useful as a pointer to the buffers. In the future, it could be used for other things. For example, we could build a feature that would ensure no duplicate `id` attributes were used (for accessibility), using this context object.

willcosgrove:
Ah ok, that makes total sense. I guess you could say I was missing that piece of context 🥁 

Let me update this real quick...

willcosgrove:
Ok, for the purposes of making this fix as minimally invasive as possible, I've skipped over the renaming here. Maybe it's something we can circle back on for a 2.0 change, because I do think it makes the internals a little easier to grok.

This PR is basically exactly what you proposed earlier, except that I've combined `capturing` and `with_target` into one method `capturing_into(target)`. Let me know if you think there's a reason those two steps need to be kept separate.

joeldrapper (approved):

PR #568, opened 2023-04-22
Minor fixes to article usage in element documentation

willcosgrove:


lib/phlex/svg/standard_elements.rb
```diff
@@ -74,175 +74,175 @@ module Phlex::SVG::StandardElements
 	register_element :ellipse, tag: "ellipse"
 
 	# @!method feBlend(**attributes, &content)
-	# 	Outputs a `<feBlend>` tag.
+	# 	Outputs an `<feBlend>` tag.
```
joeldrapper, line 77: I guess this is pronounced `eff-ee-blend` rather than `fee-blend`. Good catch.

willcosgrove: I've never used it, so I could be wrong 🤷‍♂️ but I would have naively read it "eff-ee"

PR #570, opened 2023-04-27
Add `canvas` tag to the list of standard HTML elements

blocknotes:
This PR adds the `canvas` element to the list of standard HTML elements.

Ref: https://developer.mozilla.org/docs/Web/HTML/Element/canvas


joeldrapper:
Thank you.

PR #573, opened 2023-05-02
Begin switching to GreenDots

joeldrapper:


.github/workflows/ci.yml
```diff
@@ -39,6 +39,9 @@ jobs:
       - name: Tests
         run: bundle exec sus
 
+      - name: GreenDots tests
+        run: bundle exec gd gd
```
joeldrapper, line 43: We need the extra `gd` here for the test directory, since `test` is taken by Sus.

PR #611, opened 2023-12-08
Use `yield_content` in `whitespace`

willcosgrove:
This will allow the block form of `whitespace` to work the same way you would work with an element method. Your block can return a string, or any object that can be formatted by `format_object` if it contains no other element methods.

```rb
span { "9" }
whitespace { "out of" }
span { "10" }
```


joeldrapper:
Good idea! 🙌

bradgessler:
Dig the simplicity of this. I had a whole "we need to have a join method that handles spaces" ... this is way better.

Issue #613, opened 2023-12-14
`mix` helper does not intuitively mix different types


willcosgrove:
When I added support for Array/Set attribute values in #546, I did not consider how it would impact the `mix` helper.

Currently, if the old and new values being mixed together are not the same type, the new value always replaces the old value. Now that attribute values can be Arrays or Sets, it seems like it might be a good idea for `mix` to handle merging across types.

For instance if I have some component like a button that is going to define some basic classes and then mix in any user provided attributes, and the user wants to provide more classes using an array, it shouldn't lose all the base classes.

```rb
class Button < Phlex::HTML
  def initialize(**attrs)
    @attrs = attrs
  end

  def template(&)
    button(**mix({class: "bg-gray-200 text-gray-900 text-sm"}, @attrs), &)
  end
end

render Button.new(class: ["w-full", ("bg-red-100 text-red-700" if errors.any?)])
```

With the current implementation of `mix`, we lose the classes defined in the component. Ideally I think they would combine like they were all defined as arrays.

PR #620, opened 2024-01-09
Add support for using the `mix` helper with different types

willcosgrove:
See #613 for examples where this is necessary.

I'm not in love with the approach here, but the double case was what I found easiest to reason about. If you have any suggestions for improvement let me know.


joeldrapper:
The nested case looks good to me.

joeldrapper (approved):
Looks like there's a Rubocop issue here. Otherwise, this is good to go.

joeldrapper:
I re-ran the rubocop action and it passed, so I’ll ship this now.

PR #621, opened 2024-01-20
Move the Helpers include to parent class so Phlex::SVG subclasses have access

davekaro:
I was trying to use `mix` in a Phlex::SVG object but got no method error.


joeldrapper:
Thanks @davekaro

PR #630, opened 2024-01-30
Rename `template` → `view_template`

joeldrapper:
This is a non-breaking rename that issues a warning whenever you define `template`.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -55,22 +55,32 @@ module Phlex
 
 		# @abstract Override to define a template for your component.
 		# @example
-		# 	def template
+		# 	def view_template
 		# 		h1 { "👋 Hello World!" }
 		# 	end
 		# @example Your template may yield a content block.
-		# 	def template
+		# 	def view_template
 		# 		main {
 		# 			h1 { "Hello World" }
 		# 			yield
 		# 		}
 		# 	end
 		# @example Alternatively, you can delegate the content block to an element.
-		# 	def template(&block)
+		# 	def view_template(&block)
 		# 		article(class: "card", &block)
 		# 	end
 		def template
 			yield
+		end
+
+		def self.method_added(method_name)
+			if method_name == :template
+				Kernel.warn "Defining the `template` method on a Phlex component is deprecated and will be unsupported in Phlex 2.0. Please define `view_template` instead."
+			end
+		end
+
+		def view_template(&)
+			template(&)
 		end
 
 		# @api private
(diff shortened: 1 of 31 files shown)


joeldrapper:
Because there’s an HTML tag `<template>`. In `2.0`, the `template` method will render a `<template>` tag. In `1.10`, you’ll be able to define a component template with either `template` or `view_template`, but using `template` is deprecated.

davidalejandroaguilar:
@joeldrapper Thanks for your response! Funny enough, I had just found that article myself right before you got back to me.

So, choosing to rename it to `view_template` turned out to be a better call than simply sticking with `template` and having that HTML tag be the only one called via `template_tag`, huh? 

I get the urge to make that tweak as well – my "OCD" would have nudged me in the same direction. But I can't help thinking that using the shorter `template` instead of `view_template` would have been more up my alley.

davidalejandroaguilar:
Absolutely, that makes sense. I'm curious, do you see this change as a lasting one? Or do you think the compiler improvement you talked about in the article might allow us to revert back to using `template` in the future?

Also, I really appreciate you taking the time to break this down for me. I hope it didn't pull you away from your work for too long. It would be great if the reasoning behind changes like these could be included in the PR descriptions for those of us keeping an eye on the project.

joeldrapper:
For now, the compiler is a really low priority because Phlex is so fast already. I expect `view_template` is here to stay now.

lib/phlex/sgml.rb
```diff
@@ -55,24 +55,34 @@ def element_method?(method_name)
...
+		# 	def view_template(&block)
 		# 		article(class: "card", &block)
 		# 	end
 		def template
 			yield
 		end
 
+		def self.method_added(method_name)
+			if method_name == :template
+				Kernel.warn "Defining the `template` method on a Phlex component is deprecated and will be unsupported in Phlex 2.0. Please define `view_template` instead."
```
simi, line 78: :bulb: Seems I'm late to party with my story, since this check was already removed. I had to spent some time to find out which component emits this warning. If possible, add also some pointer like `self.class` to the output for future warning. :pray: I was able to find out just by opening gem with bundle open and adding debug print in this method temporarily in the end.

joeldrapper: Hey @simi, thanks for the feedback. Yes, we removed it from `main`, but it’s still in the latest version. I’ll try to remember to add more context to deprecation warnings in the future.

I hope we won’t need to do anymore of these going forward.

PR #635, opened 2024-01-30
Remove ERB dependency

joeldrapper:
ERB is bundled with Ruby already

### DIFF
--- a/lib/phlex.rb
+++ b/lib/phlex.rb
@@ -10,6 +10,14 @@ module Phlex
 	#  rescue Phlex::Error
 	module Error; end
 
+	if defined?(ERB::Escape)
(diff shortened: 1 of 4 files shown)


lib/phlex.rb
```diff
@@ -9,6 +9,14 @@ module Phlex
 	#  rescue Phlex::Error
 	module Error; end
 
+	if defined?(ERB::Escape)
+		Escape = ERB::Escape
+	elsif defined?(ERB::Util)
+		Escape = ERB::Util
+	else
+		Escape = BasicEscape
```
joeldrapper, line 17: I ended up removing this because you need to require ERB anyway, and all versions of ERB have `ERB::Util`.

PR #639, opened 2024-01-30
Simplify calls to `register_element`

brandondrew:
This removes redundancy in calling `register_element`, since the `tag:` parameter is _almost always_ unnecessary, as it can be computed based on the `method_name`.

I'm sorry that I haven't fully tested this change because I wasn't able to get the green dots tests working... filing a bug for that shortly...


joeldrapper:
The change is already tested because our tests use the list of registered_elements.

PR #644, opened 2024-01-31
Public Context

willcosgrove:
I wanted to open this draft PR so we had a space to talk about the API here.

* Internally I've called this data `user_context`
* The public API to access it is just at `context`
* To allow for setting `user_context` through the `call` method, I've needed to rename the existing `context` keyword argument. So I made both it and `parent` use our internal naming scheme `_context:, _parent:`
* `context` is by default an empty Hash

What are everyone's thoughts?

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -2,12 +2,13 @@
 
 # @api private
 class Phlex::Context
-	def initialize
+	def initialize(user_context = {})
 		@target = +""
 		@capturing = false
+		@user_context = user_context
 	end
 
-	attr_accessor :target, :capturing
+	attr_accessor :target, :capturing, :user_context
 
 	def capturing_into(new_target)
 		original_target = @target
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -102,10 +102,10 @@ module Phlex
 		end
 
 		# @api private
-		def __final_call__(buffer = +"", context: Phlex::Context.new, parent: nil, &block)
+		def __final_call__(buffer = +"", context: {}, _context: Phlex::Context.new(context), _parent: nil, &block)
 			@_buffer = buffer
-			@_context = context
-			@_parent = parent
+			@_context = _context
+			@_parent = _parent
 
 			block ||= @_content_block
 
@@ -130,7 +130,13 @@ module Phlex
 				end
 			end
 
-			buffer << context.target unless parent
+			buffer << _context.target unless _parent
+		end
+
+		# Access the current render context data
+		# @return the supplied context object, by default a Hash
+		def context
+			@_context.user_context
 		end
 
 		# Output text content. The text will be HTML-escaped.
@@ -222,10 +228,10 @@ module Phlex
 		def render(renderable, &block)
 			case renderable
 			when Phlex::SGML
-				renderable.call(@_buffer, context: @_context, parent: self, &block)
+				renderable.call(@_buffer, _context: @_context, _parent: self, &block)
 			when Class
 				if renderable < Phlex::SGML
-					renderable.new.call(@_buffer, context: @_context, parent: self, &block)
+					renderable.new.call(@_buffer, _context: @_context, _parent: self, &block)
 				end
 			when Enumerable
 				renderable.each { |r| render(r, &block) }
--- a/test/phlex/view/capture.rb
+++ b/test/phlex/view/capture.rb
@@ -84,20 +84,15 @@ describe Phlex::HTML do
 		view do
 			def view_template
 				h1 { "Before" }
-				render @_context.view_context.previewer do
-					render @_context.view_context.component
+				render context[:test_scope].previewer do
+					render context[:test_scope].component
 				end
 				h1 { "After" }
 			end
 		end
 
 		let(:context) do
-			Phlex::Context.new.tap do |context|
-				class << context
-					attr_accessor :view_context
-				end
-				context.view_context = self
-			end
+			{ test_scope: self }
 		end
 
 		it "should contain the full capture" do
--- a/test/phlex/view/context.rb
+++ b/test/phlex/view/context.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+describe Phlex::HTML do
+	extend ViewHelper
+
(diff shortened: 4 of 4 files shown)


joelmoss:
I'm a little unsure as to the use case here. I've looked at the tests but I'm still not sure. Could you perhaps explain the problem and how this solves it? thx 🙏

joelmoss:
Oooh OK. I think I like this.

How would you pass context to a component? Would it be part of it's initializer?

willcosgrove:
As implemented here, it gets passed in at render time, not instantiation time. Specifically in the `call` method.

So the Phlex API would just be to allow you to pass the context when you use `call`:
```ruby
MyComponent.call(context: { ... })
```

In practice, `call` usually gets called by `phlex-rails` inside the `render_in` definition. I think we would expose this so that in your Rails controller you could define a method, maybe call it `phlex_context`, and `render_in` would automatically inject it for you if it was defined. Then you can define it in an `ApplicationController` or further down the hierarchy.

```ruby
class ApplicationController < ActionController::Base
  def phlex_context = {
    breadcrumbs: [{name: "My App", url: "/"}]
  }
end

class PostsController < ApplicationController
  def show
    @post = Post.find(params[:id])
    render Posts::ShowView.new(@post)
  end

  # ok this is a little gnarly now that I'm writing it out. I would make a breadcrumb helper macro in real life
  def phlex_context
    super.merge({
      breadcrumbs: [
        { name: "Posts", url: posts_path },
        ({ name: @post.name, url: post_path(@post) } if @post),
      ].compact
    }) do |key, old_value, new_value|
      if key == :breadcrumbs
        old_value + new_value
      else
        new_value
      end
    end
  end
end
```

There's a lot there that I'm making up on the fly, and all of that is debatable. But _somehow_ context would be able to be passed to a component being rendered. So this PR is mostly about the first half of that equation, of accepting `context`. `phlex-rails` will need its own PR to utilize `context`.

joelmoss:
You see now this is where it becomes very similar to defining instance variables in the controller, and having them available in the views.

```ruby
class ApplicationController < ActionController::Base
  def phlex_context = {
    breadcrumbs: [{name: "My App", url: "/"}]
  }
end
```

vs

```ruby
class ApplicationController < ActionController::Base
  def show
    @breadcrumbs = [{name: "My App", url: "/"}]
  end
end
```

So as far as Rails usage is concerned, I'm seeing little value to the context idea.

joeldrapper:
To me, passing a context object is quite different from implicitly copying down instance variables indiscriminately from the controller. Context is explicit, and designed for data that’s always expected to be available in a large number of views (if not all the views). You’d never use it to pass a `post` to the `Posts::ShowView`, but you might use it to pass the currently logged in user.

joelmoss:
But based on the example that @willcosgrove gave, the context isn't actually being passed anywhere. It's just being defined within a method. It's implicitly given to the views, just like instance vars are.

```ruby
class ApplicationController < ActionController::Base
  def phlex_context = {
    breadcrumbs: [{name: "My App", url: "/"}]
  }
end
```

So unless we get explicit and pass the context directly to the view, I see no benefit here,

```ruby
class ApplicationController < ActionController::Base
  def show
    render MyView.new breadcrumbs: [{name: "My App", url: "/"}]
  end
end
```

willcosgrove:
> I see no benefit here

The explicitness or implicitness of how it is delivered is irrelevant to the benefit, in my opinion.

The benefit is that context is shared through the whole tree, whereas an argument passed to the view will only be accessible in that one view, unless it also passes it down.

joelmoss:
I think we misunderstand each other.

> The benefit is that context is shared through the whole tree, whereas an argument passed to the view will only be accessible in that one view, unless it also passes it down.

Yes this is of course a clear benefit, and is why I like it.

But my feedback is based solely on the way that the context is actually defined and how it gets to the views from the controller.

So as you suggested - and correct me if I got this wrong - this would be achieved by defining a `phlex_context` method, and Phlex would automatically expose/copy that to the views.

```ruby
class ApplicationController < ActionController::Base
  def phlex_context = {
    breadcrumbs: [{name: "My App", url: "/"}]
  }
end
```

But my point is that this is very similar to the way Rails automatically exposes/copies instance vars from the controller to the views.

```ruby
class ApplicationController < ActionController::Base
  def show
    @breadcrumbs = [{name: "My App", url: "/"}]
  end
end
```

Hope I'm communicating things clearly. Apologies if not 😏

willcosgrove:
Gotcha, I understand. My confusion was because this PR isn't actually defining anything about _how_ the context will be delivered, just _that it can be_.

We will need to open a PR on `phlex-rails` with an API design for how context gets passed into the views. My `phlex_context` was just one suggestion. I know everyone has differing opinions on what level of implicitness/explicitness is best.

Just to know where you stand, are you wanting it to be implicit (like how Rails handles ivars) or are you wanting it to be explicitly passed in by the user? Or are you making a comment more broadly that if we're doing this one thing implicitly, then we should do all ivars implicitly?

joelmoss:
> We will need to open a PR on phlex-rails with an API design for how context gets passed into the views. My phlex_context was just one suggestion. I know everyone has differing opinions on what level of implicitness/explicitness is best.

👌

> Just to know where you stand, are you wanting it to be implicit (like how Rails handles ivars) or are you wanting it to be explicitly passed in by the user? Or are you making a comment more broadly that if we're doing this one thing implicitly, then we should do all ivars implicitly?

First, new users to Phlex may expect the Rails ivars implicity, in the same way that they expect all the helpers to be available. So we should consider that.

Secondly, I personally favour explicitness, but still in the Railsy way. For example, this is how I currently do it using my [Phexible](https://github.com/joelmoss/phlexible/tree/controller_vars?tab=readme-ov-file#controllervariables) gem...

```ruby
class Views::Users::Index < Views::Base
  include Phlexible::Rails::ControllerVariables

  controller_variable :first_name, :last_name

  def template
    h1 { "#{@first_name} #{@last_name}" }
  end
end
```

Any instance variable defined in the controller can be exposed to any Phlex view by declaring `controller_variable`. This still works in an explicit way, but also allows me to access any ivar in any view.

It's a similar concept to what we're talking about, but the difference is how and where the context or variable is defined.

PR #647, opened 2024-02-02
Add an LRU cache

joeldrapper:


lib/phlex/lru.rb
```diff
@@ -0,0 +1,69 @@
...
+	end
+
+	def fetch(key)
+		@mutex.synchronize do
+			if (value = @hash[key])
+				value
+			else
+				@hash[key] = yield
+			end
+		end
```
joeldrapper, line 53: This isn't quite right. Needs to mirror the `[]=` method above.

Issue #632, opened 2024-01-30
Remove the `<param>` element


joeldrapper:
- [ ] Issue a warning in 1.x
- [ ] Remove in 2.0

stephannv:
Solved by: https://github.com/phlex-ruby/phlex/pull/642/files

joeldrapper:
@stephannv That `deprecated: true` argument is just a no-op at the moment. We need to update the generated method to issue a warning.

PR #649, opened 2024-02-02
Warn about deprecated element methods

stephannv:
Fix first item from https://github.com/phlex-ruby/phlex/issues/632

### DIFF
--- a/fixtures/view_helper.rb
+++ b/fixtures/view_helper.rb
@@ -2,6 +2,8 @@
 
 module ViewHelper
 	def self.extended(parent)
+		parent.include ViewHelper::Utils
+
 		parent.class_exec do
 			let(:output) { example.call }
 			let(:example) { view.new }
@@ -17,6 +19,20 @@ module ViewHelper
 	def svg_view(&block)
 		let :view do
 			Class.new(Phlex::SVG, &block)
+		end
+	end
+
+	module Utils
+		def capture_stderr(&block)
+			captured_stream = StringIO.new
+			original_stream = $stderr
+			$stderr = captured_stream
+
+			block.call
+
+			captured_stream.string.strip
+		ensure
+			$stderr = original_stream
 		end
 	end
 end
--- a/lib/phlex/elements.rb
+++ b/lib/phlex/elements.rb
@@ -35,6 +35,10 @@ module Phlex::Elements
 			# frozen_string_literal: true
 
 			def #{method_name}(**attributes, &block)
+				if #{deprecated}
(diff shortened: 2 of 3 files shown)


fixtures/view_helper.rb
```diff
@@ -19,4 +21,18 @@ def svg_view(&block)
...
+			captured_stream = StringIO.new
+			original_stream = $stderr
+			$stderr = captured_stream
+
+			block.call
+
+			captured_stream.string.strip
+		ensure
+			$stderr = original_stream
+		end
```
stephannv, line 36: I added this as a module because I think it can be useful for other deprecation tests in the future.

lib/phlex/elements.rb
```diff
@@ -35,6 +35,10 @@ def register_element(method_name, tag: nil, deprecated: false)
 			# frozen_string_literal: true
 
 			def #{method_name}(**attributes, &block)
+				if #{deprecated}
+					Kernel.warn("`#{method_name}` and `_#{method_name}` are deprecated and will be unsupported in Phlex 2.0. This HTML element is no longer recommended. Check out MDN web docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/#{tag}.")
```
stephannv, line 39: 1) I couldn't get called method name here (eg. `param` or `_param`), using `__method__` or `__callee` returns `register_element` here or `register_void_element` on line 82.

2) I'm not sure if it makes sense to add link to MDN docs, but it explains why the method is being deprecated and redirects the user to more detailed context.

test/phlex/view/tags.rb
```diff
@@ -122,4 +122,40 @@
 			end
 		end
 	end
+
+	describe "Elements deprecation" do
+		deprecated_methods = %i[param]
+
+		deprecated_methods.each do |method_name|
```
stephannv, line 129: I'm not sure if worths to test non deprecated elements. I would be something like:
```ruby
[
	Phlex::HTML::StandardElements.registered_elements,
	Phlex::HTML::VoidElements.registered_elements
].each do |element_group|
	element_group.each do |method_name, tag|
    if deprecated_methods.include?(method_name)
      # tests for method_name and _method_name
      ... warns about deprecation
    else
      # tests for method_name and _method_name
      ... doesn't warn about deprecation
    end
  end
end
```

joeldrapper:
Sorry @stephannv, I wanted to go with a simpler solution to this where the message was just passed to the `deprecated:` argument. See #658 

I really appreciate the effort you put into this PR, but it didn’t feel right for what this is, especially as this is probably the only time we'll have to deprecate elements.

stephannv:
No problem @joeldrapper.

PR #655, opened 2024-02-09
Phlex::CSV

joeldrapper:
In addition to HTML and SVG, Phlex will soon support CSVs.

This is all based on @willcosgrove’s explorations here https://gist.github.com/willcosgrove/004389a5a053bccb35016223317cb9e9

### DIFF
--- a/lib/phlex/csv.rb
+++ b/lib/phlex/csv.rb
@@ -43,8 +43,8 @@ class Phlex::CSV
 	def column(header = nil, value)
 		if @_first
 			@_headers << header
-		else
-			raise unless header == @_headers[@_current_column_index]
+		elsif header != @_headers[@_current_column_index]
+			raise "Inconsistent header."
 		end
 
 		@_current_row << value
--- a/lib/phlex/csv.rb
+++ b/lib/phlex/csv.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "csv"
-
 class Phlex::CSV
 	include Phlex::Callable
 
@@ -22,10 +20,10 @@ class Phlex::CSV
 				template(*args, **kwargs)
 
 				if @_first && render_headers?
-					buffer << ::CSV.generate_line(@_headers)
+					buffer << @_headers.map! { |value| escape(value) }.join(",") << "\n"
 				end
 
-				buffer << ::CSV.generate_line(@_current_row)
+				buffer << @_current_row.map! { |value| escape(value) }.join(",") << "\n"
 				@_current_column_index = 0
 				@_current_row.clear
 			end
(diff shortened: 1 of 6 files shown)


test/phlex/csv.rb
```diff
@@ -0,0 +1,30 @@
+class Report < Phlex::CSV
+	def template(user)
+		user => { id:, name:, email: }
```
joeldrapper, line 3: I just wanted to try this lol

lib/phlex/csv.rb
```diff
@@ -0,0 +1,75 @@
...
+			raise unless header == @_headers[@_current_column_index]
+		end
+
+		@_current_row << value
+		@_current_column_index += 1
+	end
+
+	def collection_yielder(record)
+		yield(record)
+	end
```
joeldrapper, line 58: This is nice. I wonder if it would be useful to have another hook for enumerating the collection.

```
def each_item(collection, &)
  collection.each(&)
end
```

lib/phlex/csv.rb
```diff
@@ -0,0 +1,75 @@
...
+
+	private
+
+	def column(header = nil, value)
+		if @_first
+			@_headers << header
+		else
+			raise unless header == @_headers[@_current_column_index]
+		end
```
joeldrapper, line 50: I moved this `@_first` check up to an ivar to avoid the method rename.

lib/phlex/csv.rb
```diff
@@ -0,0 +1,75 @@
...
+	end
+
+	def render_headers?
+		true
+	end
+
+	def helpers
+		@_view_context
+	end
+
```
joeldrapper, line 71: Missing a reader method for the collection. Must have cut that somehow.

test/phlex/csv.rb
```diff
@@ -0,0 +1,30 @@
...
+
+		column("id", id)
+		column("name", name)
+		column("email", email)
+	end
+end
+
+User = Data.define(:id, :name, :email)
+
+it "works" do
```
joeldrapper, line 13: These tests are garbage

lib/phlex/csv.rb
```diff
@@ -0,0 +1,75 @@
...
+		nil
+	end
+
+	private
+
+	def column(header = nil, value)
+		if @_first
+			@_headers << header
+		else
+			raise unless header == @_headers[@_current_column_index]
```
joeldrapper, line 49: Should raise a proper error with a description, etc. here.

lib/phlex/csv.rb
willcosgrove: This is probably perfectly reasonable. I think I did it in the name of performance. There was something frustrating about having a branching conditional that I knew would only be true one time.

PR #660, opened 2024-02-19
Extract `Phlex::Renderable`

joeldrapper:


joeldrapper:
I'm debating whether `content_type` and `file_name` should be part of this interface. 🤔

lib/phlex/sgml.rb
```diff
@@ -222,10 +223,10 @@ def flush
 		# 		render @items
 		def render(renderable, &block)
 			case renderable
-			when Phlex::SGML
+			when Phlex::Renderable
```
joeldrapper, line 226: One problem with this is you could potentially render an unsafe CSV into an HTML view.

joeldrapper: Actually, that would prevent you from rendering SVGs inside HTMLs. Perhaps it could be guarded by `html_safe?`/`svg_safe?` on the renderable. 🤔

PR #662, opened 2024-02-19
Prevent CSV injection attacks

### LINKED ISSUE #659: Handle leading and trailing `\s` characters in CSVs
I’m leaning towards defaulting to trimming the strings by default, but providing an option to preserve leading and trailing space. Those values would need to be quoted as part of the CSV escape.

### LINKED ISSUE #661: Prevent CSV injection attacks
See: 
1. https://owasp.org/www-community/attacks/CSV_Injection
3. http://georgemauer.net/2017/10/07/csv-injection.html

joeldrapper:
Closes #661 and also closes #659.

## Todo

- [ ] Test formulae attacks
- [ ] Test trimming whitespace

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,3 +22,6 @@ Naming/AsciiIdentifiers:
 
 Naming/MethodName:
   Enabled: false
+
+Style/ReturnNilInPredicateMethodDefinition:
+  Enabled: false
--- a/lib/phlex/csv.rb
+++ b/lib/phlex/csv.rb
@@ -76,6 +76,7 @@ class Phlex::CSV
 		true
 	end
 
+	# Override and set to `true` to strip leading and trailing whitespace from values.
 	def strip_values?
 		false
 	end
@@ -97,7 +98,7 @@ class Phlex::CSV
 			# Prefix a single quote to prevent Excel, Google Docs, etc. from interpreting the value as a formula.
 			# See https://owasp.org/www-community/attacks/CSV_Injection
 			%("'#{value.gsub('"', '""')}")
-		elsif SPACE_CHARACTERS[first_char] || SPACE_CHARACTERS[last_char] || value.include?('"') || value.include?(",") || value.include?("\n")
+		elsif (!strip_values? && (SPACE_CHARACTERS[first_char] || SPACE_CHARACTERS[last_char])) || value.include?('"') || value.include?(",") || value.include?("\n")
 			%("#{value.gsub('"', '""')}")
 		else
 			value
--- a/lib/phlex/csv.rb
+++ b/lib/phlex/csv.rb
@@ -72,6 +72,7 @@ class Phlex::CSV
 		nil
 	end
 
+	# Override and set to `false` to disable rendering headers.
 	def render_headers?
 		true
 	end
@@ -81,6 +82,7 @@ class Phlex::CSV
 		false
 	end
 
+	# Override and set to `false` to disable CSV injection escapes.
 	def prevent_csv_injection?
 		true
 	end
--- a/lib/phlex/csv.rb
+++ b/lib/phlex/csv.rb
@@ -6,22 +6,6 @@ class Phlex::CSV
 	FORMULA_PREFIXES = ["=", "+", "-", "@", "\t", "\r"].to_h { |prefix| [prefix, true] }.freeze
 	SPACE_CHARACTERS = [" ", "\t", "\r"].to_h { |char| [char, true] }.freeze
 
-	module InjectionPolicy
-		Raise = lambda do |value|
-			raise "CSV injection detected: #{value.inspect}"
-		end
-
-		Ignore = lambda do |value|
-			value
-		end
-
-		Escape = lambda do |value|
-			%("'#{value.gsub('"', '""')}")
-		end
-	end
-
-	include InjectionPolicy
-
 	def initialize(collection)
 		@collection = collection
 		@_headers = []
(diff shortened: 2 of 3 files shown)


lib/phlex/csv.rb
```diff
@@ -82,9 +82,13 @@ def render(renderable)
 	end
 
 	def escape(value)
-		value = value.to_s
+		value = value.to_s.dup
+		value.strip!
```
joeldrapper, line 86: This should be optional. Perhaps you should be able to define `def strip_values? = false` or something like that.

lib/phlex/csv.rb
```diff
@@ -15,17 +18,21 @@ def initialize(collection)
...
 				view_template(*args, **kwargs)
 
 				if @_first && render_headers?
-					buffer << @_headers.map! { |value| escape(value) }.join(",") << "\n"
+					buffer << @_headers.join(",") << "\n"
 				end
 
-				buffer << @_current_row.map! { |value| escape(value) }.join(",") << "\n"
+				buffer << @_current_row.join(",") << "\n"
 				@_current_column_index = 0
```
joeldrapper, line 36: I’m moving the escape down to the `column` method.

lib/phlex/csv.rb
```diff
@@ -48,43 +55,56 @@ def content_type
...
+		@_current_row << escape(value)
 		@_current_column_index += 1
 	end
 
 	def each_item(&block)
 		collection.each(&block)
 	end
 
-	def collection_yielder(record)
+	def yielder(record)
```
joeldrapper, line 71: Unrelated to the main purpose of this PR, I think I prefer `yielder` for this method, since it's not really yielding the collection, it's really yielding a single item in the collection.

lib/phlex/csv.rb
```diff
@@ -15,17 +18,21 @@ def initialize(collection)
 	attr_reader :collection
 
 	def call(buffer = +"", view_context: nil)
+		unless prevent_csv_injection? == true || prevent_csv_injection? == false
+			raise "You must define `prevent_csv_injection?` on #{self.class.inspect}, returning `true` or `false`. See https://owasp.org/www-community/attacks/CSV_Injection"
+		end
```
joeldrapper, line 23: By settings `prevent_csv_injection?` to `nil` by default, we can force the developer to make an informed decision upfront by raising here. The developer must choose whether to enable or disable the protection, and they have an opportunity to learn about the risk on owasp.org.

.rubocop.yml
```diff
@@ -22,3 +22,6 @@ Naming/AsciiIdentifiers:
 
 Naming/MethodName:
   Enabled: false
+
+Style/ReturnNilInPredicateMethodDefinition:
+  Enabled: false
```
joeldrapper, line 27: Rubocop doesn't like the `nil` predicate `prevent_csv_injection?`.

lib/phlex/csv.rb
```diff
@@ -48,43 +55,56 @@ def content_type
...
+	def escape(value)
+		value = trim_whitespace? ? value.to_s.strip : value.to_s
+		first_char = value[0]
+		last_char = value[-1]
+
+		if prevent_csv_injection? && FORMULA_PREFIXES[first_char]
+			# Prefix a single quote to prevent Excel, Google Docs, etc. from interpreting the value as a formula.
+			# See https://owasp.org/www-community/attacks/CSV_Injection
+			%("'#{value.gsub('"', '""')}")
+		elsif (!trim_whitespace? && (SPACE_CHARACTERS[first_char] || SPACE_CHARACTERS[last_char])) || value.include?('"') || value.include?(",") || value.include?("\n")
```
joeldrapper, line 107: This line is unfortunately not particularly clear, but it's optimised for performance. For example, we don’t need to check if the first and last characters are whitespace characters if we're trimming whitespace anyway.

lib/phlex/csv.rb
```diff
@@ -3,6 +3,25 @@
 class Phlex::CSV
 	include Phlex::Callable
 
+	FORMULA_PREFIXES = ["=", "+", "-", "@", "\t", "\r"].to_h { |prefix| [prefix, true] }.freeze
+	SPACE_CHARACTERS = [" ", "\t", "\r"].to_h { |char| [char, true] }.freeze
+
+	module InjectionPolicy
+		Raise = lambda do |value|
+			raise "CSV injection detected: #{value.inspect}"
```
joeldrapper, line 11: Let's make this error really good.

lib/phlex/csv.rb
```diff
@@ -3,6 +3,25 @@
...
+		end
+
+		Ignore = lambda do |value|
+			value
```
joeldrapper, line 15: Ops, this will bypass the `else` condition in `#escape`.

lib/phlex/csv.rb
```diff
@@ -100,10 +112,8 @@ def escape(value)
 		first_char = value[0]
 		last_char = value[-1]
 
-		if prevent_csv_injection? && FORMULA_PREFIXES[first_char]
-			# Prefix a single quote to prevent Excel, Google Docs, etc. from interpreting the value as a formula.
-			# See https://owasp.org/www-community/attacks/CSV_Injection
-			%("'#{value.gsub('"', '""')}")
+		if InjectionPolicy::Ignore != in_case_of_injection && FORMULA_PREFIXES[first_char]
```
willcosgrove, line 115: Don't we need to do the injection policy after potentially trimming whitespace?

willcosgrove: Oh nevermind, I see that we've trimmed whitespace already above. I didn't have the diff expanded and I thought it was happening in the `elsif`

willcosgrove (approved):
Thanks for battling this out with me 😇

PR #670, opened 2024-03-06
Support selecting multiple fragments

joeldrapper:
This PR adds support for matching multiple fragments. Each fragment is rendered in its own `<template>` with a `data-id` attribute matching the ID of the fragment matched.

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -6,17 +6,32 @@ class Phlex::Context
 		@buffer = +""
 		@capturing = false
 		@user_context = user_context
-		@fragment = nil
+		@fragments = nil
 		@in_target_fragment = false
-		@found_target_fragment = false
 	end
 
-	attr_accessor :buffer, :capturing, :user_context, :fragment,
-		:in_target_fragment, :found_target_fragment
+	attr_accessor :buffer, :capturing, :user_context, :in_target_fragment
+
+	attr_reader :fragments
 
 	# Added for backwards compatibility with phlex-rails. We can remove this with 2.0
 	def target
 		@buffer
+	end
+
+	def target_fragments(fragments)
+		@fragments = fragments.to_h { |it| [it, +""] }
+	end
+
+	def begin_target(id)
+		@in_target_fragment = id
+		@buffer << %(<template data-id="#{id}">)
+	end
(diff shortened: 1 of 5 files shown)


lib/phlex/context.rb
```diff
@@ -6,19 +6,34 @@ def initialize(user_context = {})
...
 		@buffer
 	end
 
+	def target_fragments(fragments)
+		@fragments = fragments.to_h { |it| [it, true] }
+	end
+
+	def begin_target(id)
+		@in_target_fragment = id
+		@buffer << %(<template data-id="#{id}">)
```
joeldrapper, line 28: @midnightmonster I realised we don’t return empty `<template>`s anyway so this `data-id` isn't particularly helpful.

midnightmonster: We could, though. If we get all the way to the end and there's still something left in `@fragments`, just spit out the empty template tag.

joeldrapper: Yeah, I’m debating if this is worth it, since you can always find the missing IDs by looking at the child nodes on the client side.

midnightmonster: Remembering that there's a need for both "remove this node which is no longer present" and "silly developer, why did you ask me for a node that's inside another node you also asked for?"—I think the simplest thing is not to use the template tags at all. Client side makes the request and remembers which IDs it asked for. It instantiates the response text as a DocumentFragment (or however you're doing this) and (assuming it was requesting specific IDs at all) just goes down the list of IDs it knows it requested, querying the response document fragment by ID.

1. If it finds the ID in the response and the page, morph the page version with the new version.
2. If it finds the ID in the response but not the page, discard with perhaps a warning (presumably it does this already?).
3. If it finds the ID in the page but not the response, remove the node from the page.

This means it morphs a second time for a nested id, but those should be rare-to-never and it shouldn't result in any changes, so it probably doesn't matter.

joeldrapper: This makes sense. It’s actually a not ideal to have to manage multiple `<template>`s in the client-side code. It would be easier to make a new `<template>` and set its inner HTML to whatever came back from the server, and then iterate over the template’s child elements.

PR #671, opened 2024-03-06
Halt rendering early once all fragments have been found

willcosgrove:


lib/phlex/context.rb
```diff
@@ -23,6 +24,15 @@ def target_fragments(fragments)
 		@fragments = fragments.to_h { |it| [it, true] }
 	end
 
+	def around_render
+		return yield if @halt_signal
+
+		catch do |signal|
+			@halt_signal = signal
+			yield
+		end
```
joeldrapper, line 33: I should have mentioned that I tried this and found it slowed the benchmark significantly.

lib/phlex/sgml.rb
```diff
@@ -113,22 +113,24 @@ def __final_call__(buffer = +"", context: Phlex::Context.new, view_context: nil,
...
-					if is_a?(DeferredRender)
-						__vanish__(self, &block)
-						view_template
-					else
-						view_template do |*args|
-							if args.length > 0
-								yield_content_with_args(*args, &block)
-							else
-								yield_content(&block)
+			@_context.around_render do
```
joeldrapper, line 116: Adding another block here will likely cause significant slow downs. I wanted to use a block to implement the whole selective rendering feature so the element would call something like

```
@context.render_element(attribute[:id]) do
  rendering code here
end
```

And the context would determine whether to yield or not. The performance cost wasn't worth it.

joeldrapper:
I benchmarked this searching for a fragment really early in the page. In theory, this should have given it its best shot. Unfortunately it was slower.

Before:

```
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) +YJIT [arm64-darwin23]
                Page    68.008k i/100ms
                Page    693.068k (± 0.7%) i/s -      3.468M in   5.004713s
```

After:

```
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) +YJIT [arm64-darwin23]
                Page    59.632k i/100ms
                Page    601.793k (± 1.0%) i/s -      3.041M in   5.054132s
```

lib/phlex/context.rb
```diff
@@ -6,19 +6,45 @@ def initialize(user_context = {})
...
 	def target
 		@buffer
 	end
 
+	def target_fragments(fragments)
+		@fragments = fragments.to_h { |it| [it, true] }
+	end
+
+	def around_render
+		return yield if @halt_signal
```
joeldrapper, line 28: We should check if `!fragments || @halt_signal` here.

lib/phlex/context.rb
```diff
@@ -6,19 +6,45 @@ def initialize(user_context = {})
...
+	def begin_target(id)
+		@in_target_fragment = id
+		@buffer << %(<template data-id="#{id}">)
+	end
+
+	def end_target
+		@fragments.delete(@in_target_fragment)
+		@buffer << "</template>"
+		@in_target_fragment = false
+		throw @halt_signal if @fragments.empty?
```
joeldrapper, line 45: I believe `@fragments.length == 0` will be faster here.

lib/phlex/sgml.rb
joeldrapper: Looks like I was wrong about this.

lib/phlex/context.rb
```diff
@@ -6,19 +6,45 @@ def initialize(user_context = {})
 		@buffer = +""
 		@capturing = false
 		@user_context = user_context
-		@fragment = nil
+		@fragments = nil
 		@in_target_fragment = false
-		@found_target_fragment = false
+		@halt_signal = nil
```
joeldrapper, line 11: Nice object shape optimisation 💪

lib/phlex/context.rb
willcosgrove: What's faster? `!fragments` or `@halt_signal`? I assume `@halt_signal` is the fastest. In fact, it may faster to skip the `fragments` check entirely, since that will run on every `around_render`. It's just there to prevent a `catch` that will only happen once any way.

willcosgrove: I guess `!@fragments` would be as fast as `@halt_signal`

joeldrapper (approved):

PR #675, opened 2024-03-12
Halt early when searching for a fragment

joeldrapper:
This is my attempt to re-apply #671 after the recent changes to fragment searching. Unfortunately, the benchmarks show that this makes fragment searching slower.


joeldrapper:
Closing this as it was slower in the end.

PR #676, opened 2024-03-13
Namespaced buckets

joeldrapper:


lib/phlex/bucket.rb
```diff
@@ -4,12 +4,22 @@ module Phlex::Bucket
...
+		else
+			warn "Phlex::Bucket is experimental and may be removed from future versions of Phlex."
+		end
 
 		if Class === constant && constant < Phlex::SGML
 			define_method(name) do |*args, **kwargs, &block|
 				render(constant.new(*args, **kwargs), &block)
 			end
+
+			define_singleton_method(name) do |*args, **kwargs, &block|
```
joeldrapper, line 18: We should raise if this is already defined like we do for the instance method.

lib/phlex/sgml.rb
```diff
@@ -113,6 +113,11 @@ def __final_call__(buffer = +"", context: Phlex::Context.new, view_context: nil,
 
 			return "" unless render?
 
+			if !parent && Phlex::SUPPORTS_FIBER_STORAGE
```
joeldrapper, line 116: This `SUPPORTS_FIBER_STORAGE` check is necessary while we support Ruby < 3.2.

lib/phlex/sgml.rb
```diff
@@ -134,7 +139,12 @@ def __final_call__(buffer = +"", context: Phlex::Context.new, view_context: nil,
 				end
 			end
 
-			buffer << context.buffer unless parent
+			unless parent
+				if Phlex::SUPPORTS_FIBER_STORAGE
+					Fiber[:__phlex_component__] = original_fiber_storage
```
joeldrapper, line 144: Putting this back rather than setting it to `nil` is only necessary if you separately render a different component from scratch while rendering this component.

lib/phlex/bucket.rb
```diff
@@ -4,12 +4,22 @@ module Phlex::Bucket
...
+				Fiber[:__phlex_component__].instance_exec do
```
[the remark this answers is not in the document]
joeldrapper: It only changes this block. We could use `send(:render)` here instead.

PR #678, opened 2024-03-14
Lazy load autoloaded bucket components

joeldrapper:
Autoloaded Ruby constants will not have corresponding proxy methods on Buckets, since these methods are defined on `const_added`. This PR solves that in two ways:

### External access
Assuming a runtime environment where none of the constants are loaded by default — such as Rails with Zeitwerk during development — we use a bucket like this:

```ruby
def view_template
  Components::Heading { "Hello" }
end
```

This call will autoload the `Components` module, but it will not autoload the `Components::Heading` module, since it’s just a method call on `Components`. How we handle this is the `Components` bucket catches the call with `method_missing`, and then tries to load the constant. If loading the constant is successful, it can call the newly defined `Heading` method on `Components`.

### When included
If we include `Components` and then try to call `Heading`, again we will run into the same problem.

```ruby
class MyComponent < Phlex::HTML
  include Components

  def view_template
    Heading { "Hello" }
  end
end
```

In this case, though, we can’t use `method_missing`. If you include `Components` and `AnotherBucket` and they both define a `Heading` component, you would expect `AnotherBucket`’s `Heading` component to take precedence over `Components`’ `Heading` component. If both buckets were using the `method_missing` technique, this precedence would be based on which bucket had their component autoloaded first.

So instead, when you _include_ a bucket, we eager-load all that bucket’s components.

### DIFF
--- a/lib/phlex/bucket.rb
+++ b/lib/phlex/bucket.rb
@@ -1,23 +1,13 @@
 # frozen_string_literal: true
 
 module Phlex::Bucket
-	module Proxy
-		def method_missing(name, *args, **kwargs, &block)
-			if self.class.constants.include?(name) && self.class.const_get(name) && methods.include?(name)
-				public_send(name, *args, **kwargs, &block)
-			else
-				super
-			end
-		end
-
-		def respond_to_missing?(name, include_private = false)
-			self.class.constants.include?(name) && self.class.const_get(name) && methods.include?(name)
-		end
-	end
-
 	def self.extended(mod)
 		warn "🚨 [WARNING] Phlex::Bucket is experimental and may be removed from future versions of Phlex."
-		mod.include(Proxy)
+	end
+
+	# When a bucket is included in a module, we need to load all of its components.
+	def included(mod)
+		constants.each { |c| mod.const_get(c) if autoload?(c) }
 	end
 
 	def method_missing(name, *args, **kwargs, &block)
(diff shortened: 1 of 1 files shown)


lib/phlex/bucket.rb
```diff
@@ -5,38 +5,49 @@ def self.extended(mod)
...
-		end
-
-		constant.include(self)
-
-		if Class === constant && constant < Phlex::SGML
+		if autoload?(name)
 			define_method(name) do |*args, **kwargs, &block|
-				render(constant.new(*args, **kwargs), &block)
+				this = method(name)
+				self.class.const_get(name) # load the constant (which should override this method)
```
joeldrapper, line 15: If this constant isn’t a renderable, we should probably undef this proxy method.

lib/phlex/bucket.rb
```diff
@@ -1,24 +1,41 @@
...
+			if self.class.constants.include?(name) && self.class.const_get(name) && methods.include?(name)
+				public_send(name, *args, **kwargs, &block)
+			else
+				super
+			end
+		end
+
+		def respond_to_missing?(name, include_private = false)
+			self.class.constants.include?(name) && self.class.const_get(name) && methods.include?(name)
+		end
```
joeldrapper, line 15: I’m not 100% sure if we want these `respond_to_missing?` methods because they may slow down `respond_to?` calls.

lib/phlex/bucket.rb
```diff
@@ -1,24 +1,41 @@
...
 
 module Phlex::Bucket
+	module Proxy
+		def method_missing(name, *args, **kwargs, &block)
+			if self.class.constants.include?(name) && self.class.const_get(name) && methods.include?(name)
+				public_send(name, *args, **kwargs, &block)
+			else
+				super
+			end
+		end
```
joeldrapper, line 11: Using `method_missing` here won’t be consistent when including modules that override other modules. The method you get will be based on which was autoloaded first, not which module has higher precedence. We should probably define these methods instead.

PR #691, opened 2024-03-22
Fix for components that render content before a fragment is found

davekaro:
Fixes the issue described https://github.com/phlex-ruby/phlex-rails/issues/174#issuecomment-2015516273

I was seeing text output before a `<turbo-frame>` in my responses, and it was all from components that render the text "implicitly".

I found no performance difference with these changes.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -158,9 +158,6 @@ module Phlex
 		# @return [nil]
 		# @see #format_object
 		def plain(content)
-			context = @_context
-			return if context.fragments && !context.in_target_fragment
-
 			unless __text__(content)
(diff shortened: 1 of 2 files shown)


lib/phlex/sgml.rb
```diff
@@ -366,7 +366,7 @@ def yield_content_with_args(*args)
 
 			original_length = buffer.bytesize
 			content = yield(*args)
-			__text__(content) if original_length == buffer.bytesize
+			plain(content) if original_length == buffer.bytesize
```
joeldrapper, line 369: `plain` will raise if it’s passed something that doesn’t render, such as `nil`. But this happens all the time with implicit text. We should move the fragment guard from `plain` to `__text__` instead of switching to use `plain` instead of `__text__`.

davekaro: Done in https://github.com/phlex-ruby/phlex/pull/691/commits/f2e71930bdb058f679051d388cfb12c619135f4c.

Issue #638, opened 2024-01-30
Test for precedence of `to_s`, `to_str`, and `to_phlex_attribute_value`


joeldrapper:
_Originally posted by @willcosgrove in https://github.com/phlex-ruby/phlex/pull/629#discussion_r1471306264_

sullyvannunes:
I would like to work on this issue. 

However, I don't know exactly how to execute the tests in `gd` directory. It seems `sus` isn't the right tool to run them. Am i right?

joeldrapper:
Yes, ignore the `gd` directory, we haven’t migrated there yet. `bundle exec sus` should run the `sus` test suite.

PR #700, opened 2024-03-30
Test custom object method call precedence

sullyvannunes:
According to issue #638, it would be better to add coverage to the method call precedence of custom object attribute building.

### DIFF
--- a/test/phlex/valueable_object.rb
+++ b/test/phlex/valueable_object.rb
@@ -35,11 +35,13 @@ describe Phlex::HTML do
 	end
 
 	it "renders phlex attribute value according to method call precedence" do
-		to_phlex_attribute_value_klass = Class.new { define_method(:to_phlex_attribute_value, -> { "to_phlex_attribute_value" }) }
-		to_str_klass = Class.new(to_phlex_attribute_value_klass) { define_method(:to_str, -> { "to_str" }) }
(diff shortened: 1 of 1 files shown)


test/phlex/valueable_object.rb
```diff
@@ -0,0 +1,47 @@
...
+			div class: klass.new
+		end
+
+		expect(component.call).to be == %(<div class="to_phlex_attribute_value"></div>)
+	end
+
+	it "renders phlex attribute value according to method call precedence" do
+		to_phlex_attribute_value_klass = Class.new { define_method(:to_phlex_attribute_value, -> { "to_phlex_attribute_value" }) }
+		to_str_klass = Class.new(to_phlex_attribute_value_klass) { define_method(:to_str, -> { "to_str" }) }
+		to_s_klass = Class.new(to_str_klass) { define_method(:to_s, -> { "to_s" }) }
```
joeldrapper, line 40: I think this would need to be just one class with all three methods defined to ensure the `to_phlex_attribute` method “wins”.

joeldrapper:
Hey @sullyvannunes, thanks so much for the PR. 🙏

I left some feedback on one of the tests.

test/phlex/valueable_object.rb
sullyvannunes: Thanks for the feedback @joeldrapper!

I refactored the test so that there is only one class that defines all three methods to ensure correct method call precedence

PR #703, opened 2024-04-06
Lets go

### LINKED ISSUE #632: Remove the `<param>` element

### LINKED ISSUE #687: Dropping support for Ruby < 3.2 in 2.0
I’m considering whether we should drop support for Ruby versions below 3.2, since 3.2 is where the `Fiber[]` and `const_added` features we depend on for `Phlex::Kit` were introduced.

We could still maintain a Phlex 1.0 branch supporting Ruby >= 2.7 for a few more years. I just think it might be easier to say, “here’s the features in 2.0” without having to qualify which Ruby versions have access to them.

joeldrapper:
Closes #687 
Closes #632


.github/workflows/ci.yml
```diff
@@ -11,16 +11,11 @@ jobs:
...
-          - "3.1"
           - "3.2"
           - "3.3"
           - "head"
-          - "truffleruby-22.2"
-          - "truffleruby-22.3"
-          - "jruby-9.4.6.0"
-          - "jruby-head"
+          - "truffleruby-23.1"
+          - "truffleruby-24.0"
```
joeldrapper, line 18: These are `3.2` and `3.3` compatible.

lib/phlex.rb
```diff
@@ -42,15 +42,7 @@ class NameError < ::NameError
...
-	SUPPORTS_FIBER_STORAGE = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.2")
-end
-
-if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("3.0")
-	class Symbol
-		def name
-			to_s
-		end
-	end
+	SUPPORTS_FIBER_STORAGE = RUBY_ENGINE == "ruby"
```
joeldrapper, line 45: TruffleRuby doesn’t support `Fiber#[]` at the moment.

lib/phlex/elements.rb
```diff
@@ -98,20 +98,11 @@ def #{method_name}(**attributes, &block)
...
 
 	# @api private
-	def register_void_element(method_name, tag: method_name.name.tr("_", "-"), deprecated: false)
-		if deprecated
-			deprecation = <<~RUBY
-				Kernel.warn "#{deprecated}"
-			RUBY
-		else
-			deprecation = ""
-		end
```
joeldrapper, line 108: We're not going to need to use this for a while, so may as well get rid of it.

lib/phlex/html.rb
```diff
@@ -7,7 +7,7 @@ class HTML < SGML
...
+		EVENT_ATTRIBUTES = Set.new(%w[onabort onafterprint onbeforeprint onbeforeunload onblur oncanplay oncanplaythrough onchange onclick oncontextmenu oncopy oncuechange oncut ondblclick ondrag ondragend ondragenter ondragleave ondragover ondragstart ondrop ondurationchange onemptied onended onerror onfocus onhashchange oninput oninvalid onkeydown onkeypress onkeyup onload onloadeddata onloadedmetadata onloadstart onmessage onmousedown onmousemove onmouseout onmouseover onmouseup onmousewheel onoffline ononline onpagehide onpageshow onpaste onpause onplay onplaying onpopstate onprogress onratechange onreset onresize onscroll onsearch onseeked onseeking onselect onstalled onstorage onsubmit onsuspend ontimeupdate ontoggle onunload onvolumechange onwaiting onwheel]).freeze
```
joeldrapper, line 10: Using a Set here seems to be just as fast as a Hash of true values.

lib/phlex/html/standard_elements.rb
```diff
@@ -639,7 +639,7 @@ module Phlex::HTML::StandardElements
 	# 	@return [nil]
 	# 	@yieldparam component [self]
 	# 	@see https://developer.mozilla.org/docs/Web/HTML/Element/template
-	register_element :template_tag, tag: "template"
+	register_element :template
```
joeldrapper, line 642: Finally 😅

PR #710, opened 2024-04-25
Exclude quickdraw tests from releases

djfpaagman:
I was checking the structure of the gemspec, and noticed that it does not have the quickdraw test files in the regex for files that are excluded. 

I did some additional digging and it seems like we could also ignore these:

* fixtures/
* config/
* bench.rb

Let me know what you think, happy to add a commit to add those.


joeldrapper (approved):

joeldrapper:
Thank you!

PR #715, opened 2024-05-02
Support for boolean attributes

GyozaGuy:
A neat feature of HTML is being able to use attributes without values to specify functionality:

```html
<button type="button" disabled>Disabled button</button>
<input type="text" required>
```

I thought it might be neat if Phlex supported something similar by allowing positional attributes to act as boolean attributes. The positional attributes are merged with the keyword attributes with a value of `true`, which looks like:

```ruby
button(:disabled, type: 'button') { 'Disabled button' }
# Output: <button type="button" disabled>Disabled button</button>

input(:required, type: 'text')
# Output: <input type="text" required>

div(:any, :symbol, :or, 'string', :can, :go, :here) { 'Content' }
# Output: <div any symbol or string can go here>Content</div>
```

This is particularly useful when using custom element libraries that make extensive use of boolean attributes, such as [Shoelace](https://shoelace.style):

```rb
# Before:
sl_button(disabled: true, loading: true, outline: true, pill: true) { 'Click me' }

# After:
sl_button(:disabled, :loading, :outline, :pill) { 'Click me' }
```

### DIFF
--- a/lib/phlex/elements.rb
+++ b/lib/phlex/elements.rb
@@ -32,11 +32,12 @@ module Phlex::Elements
 		class_eval(<<-RUBY, __FILE__, __LINE__ + 1)
 			# frozen_string_literal: true
 
-			def #{method_name}(**attributes, &block)
+			def #{method_name}(*boolean_attributes, **attributes, &block)
 				context = @_context
 				buffer = context.buffer
 				fragment = context.fragments
 				target_found = false
+				attributes.merge!(Hash[boolean_attributes.map { |attribute| [attribute, true] }]) if boolean_attributes.any?
 
 				if fragment
 					return if fragment.length == 0 # we found all our fragments already
--- a/quickdraw/html/boolean_attributes.test.rb
+++ b/quickdraw/html/boolean_attributes.test.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class BooleanAttribute < Phlex::HTML
+	def view_template
+		button(:disabled, type: "button") { "Disabled button" }
+	end
+end
+
+class BooleanAndStandardAttribute < Phlex::HTML
+	def view_template
+		button(:disabled, disabled: true, type: "button") { "Disabled button" }
+	end
+end
+
+test "boolean attributes are supported" do
+	expect(BooleanAttribute.call) == "<button type=\"button\" disabled>Disabled button</button>"
+end
+
+test "using boolean attributes and standard attributes do not conflict" do
+	expect(BooleanAndStandardAttribute.call) == "<button disabled type=\"button\">Disabled button</button>"
+end


stephannv:
I think this is already implemented on Phlex (https://www.phlex.fun/).

Screenshot 2024-05-02 at 2 32 53 PM

joeldrapper:
Hey, thanks for the PR. ❤️

The only thing that gives me pause here is that you would need to remember to put all your boolean attributes before any keyword arguments. We do already support boolean attributes via setting keyword arguments to `true` or `false`.

I’m going to take a little time to think about this further. Something like this would also prevent us from using positional arguments for anything else in the future. We’ve considered using the first positional argument for text content. We've also considered using positional arguments as a way to mix attribute hashes (see the `mix` helper).

GyozaGuy:
> Hey, thanks for the PR. ❤️
> 
> The only thing that gives me pause here is that you would need to remember to put all your boolean attributes before any keyword arguments. We do already support boolean attributes via setting keyword arguments to `true` or `false`.
> 
> I’m going to take a little time to think about this further. Something like this would also prevent us from using positional arguments for anything else in the future. We’ve considered using the first positional argument for text content. We've also considered using positional arguments as a way to mix attribute hashes (see the `mix` helper).

Yeah I did wonder if there were other things being considered for the positional arguments. I just found myself using `checked: true` or `required: true` so often I wondered if there would be a more concise way to do it. Thanks for your consideration!

joeldrapper:
> Yeah I did wonder if there were other things being considered for the positional arguments.

Yeah, I think it would be a non-breaking change to merge this, but then we couldn’t remove it or use them for something else in the future.

> I just found myself using checked: true or required: true so often I wondered if there would be a more concise way to do it. 

Yeah, this makes sense, especially if you’re applying those attributes directly. When I use attributes like `selected`, `disabled`, etc. I’m using referencing some other variable or calling some other method, so it makes sense in that case to pass a boolean.

I’ll definitely think more about this. I’m interested in @willcosgrove’s thoughts too.

willcosgrove:
Hi @GyozaGuy, thanks for the PR! It is an interesting idea.

My first thoughts are that I have not yet encountered a time where I've needed to use lots of constant boolean attributes in my HTML. It would definitely be the exception, rather than the norm that I set a boolean HTML attribute to a constant value.

```rb
# Always disabled no matter what
input(disabled: true)

# vs disabled based on context
input(disabled: !@policy.can_edit?)
```

The `required` attribute in your example may be the only exception to that, where I probably do use that specific attribute most as a constant rather than conditionally.

I can see how an API design, like Shoelace has, lends itself to using lots of boolean attributes as constant values. However I'm not sure that it carries over to being a common situation for normal HTML elements.

I think I would be much more likely to use some of the other usecases @joeldrapper outlined that has been considered for the positional arguments. So my personal vote would be to not use it for this.

---

If I were using shoelace and the attribute lists were difficult to look at/work with. I would probably make a shoelace specific wrapper.

I don't know how you've setup shoelace in your project, but I imagine somewhere you're doing something like this:

```rb
register_element :sl_alert
register_element :sl_badge
register_element :sl_button
# etc
```

I would probably define a `register_shoelace_element` that would wrap `register_element`s generated method.

```rb
def self.register_shoelace_element(method_name)
  register_element(method_name)
  element_wrapper = Module.new
  element_wrapper.module_eval(<<~RUBY)
    def #{method_name}(*boolean_attributes, **attributes, &content)
      boolean_attributes.each do |key|
        attributes[key] = true
      end

      super(**attributes, &content)
    end
  RUBY
  prepend element_wrapper
end

register_shoelace_element :sl_alert
register_shoelace_element :sl_badge
register_shoelace_element :sl_button
```

I'm sure there's plenty of room for improvement in that snippet, but that's the basic idea.

Another possible idea is to make a `shoelace` or `sl` method, which would take the shoelace component name as the first positional argument, and any subsequent positional arguments as boolean attributes.

```rb
def sl(component, *boolean_attributes, **attributes, &content)
  # merge boolean attributes ...
  send("sl_#{component}", **attributes, &content)
end

sl(:button, :pill, :loading)
```

dhnaranjo:
If this feature is merged it also needs to be documented. Now the documentation for arguments is something like:

> You provide arguments to your tags as keyword arguments:
> 
> code example here
> 
> Or you can, for attributes that will always evaluate to true, you can provide them as symbol or string positional arguments:
> 
> additional code example here


Referring to the existing behavior in html explains it effectively but doesn't really stand alone, it depends on that existing knowledge. This also raises the question of if this specific affordance for always-on attributes exists then is it the more appropriate way to provide them, or is it a style choice?

I'd prefer this weren't included in the base library, but I do think @willcosgrove's suggestion for how to implement it yourself is a great example of the whole "Phlex, it's Ruby!" Thing. The folks at https://avohq.io maintain a "cookbook" section of their docs with little like, ad hoc examples of people solving problems local extension of the library, rather than changing the library itself. Maybe this is a good candidate for something like that?

GyozaGuy:
Thanks for the comments! I'm already using a wrapper similar to @willcosgrove's suggestion for my own components, I like the idea of doing the same for Shoelace components. After consideration I agree this isn't likely to be a common scenario for most HTML and I probably only thought of this because I'm using Shoelace. I'm happy with implementing this for my own projects as needed.

joeldrapper:
Thanks @GyozaGuy. Let’s close this for now and keep our options open. ❤️

PR #717, opened 2024-05-13
Allow binding to  via the  kwarg in tokens

bradgessler:
This commit provides a shortcut for binding to a `class` keyword argument and passing it into tokens, like this:

```ruby
def view_template
  red(class: "buzz")
end

def red(class:)
  div(class: tokens("fizz", class:))
end
```

Prior to this commit, I'd have to manually bind to the `class` keyword argument passed into tokens like this:

```ruby
def red(class:)
  div(class: tokens("fizz", binding.local_variable_get(:class)))
end
```

### DIFF
--- a/lib/phlex/helpers.rb
+++ b/lib/phlex/helpers.rb
@@ -18,7 +18,7 @@ module Phlex::Helpers
 	# 	tokens(
 	# 		active?: { then: "active", else: "inactive" }
 	# 	)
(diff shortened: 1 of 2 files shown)


joeldrapper:
This is an awesome idea. I want to do something with it, but I don’t think `tokens` is the right place for this as it is currently implemented. Keyword arguments on the `tokens` method are for conditional tokens, where the key is the condition and the value is the token set.

PR #719, opened 2024-05-15
Add Phlex::SGML::SafeValue to require people to mark attribute values as 'safe' for use in attributes

bradgessler:
This PR will make it possible to mark values as "safe" to use within "dangerous" Phlex attributes like `onclick`, etc.

It accomplishes this by wrapping a value in a `Phlex::SGML::SafeValue.new("alert('1');")` class and passing it through the Phlex attribute builder. Phlex then detects this class type and passes the string through to the attribute, unescaped.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -5,6 +5,8 @@ module Phlex
 	class SGML
 		include Helpers
 
+		autoload :SafeValue, "phlex/sgml/safe_value"
+
 		class << self
 			# Render the view to a String. Arguments are delegated to {.new}.
 			def call(...)
@@ -182,6 +184,13 @@ module Phlex
 
 			@_context.target << content
 			nil
+		end
+
+		# Marks strings as "safe" so they don't raise an exception when JavaScript or potentially dangerous values are passed into Phlex attributes.
+		# @param content [String|nil]
+		# @return [SafeValue]
+		def safe(value)
+			SafeValue.new(value)
 		end
 
 		# Capture a block of output as a String.
@@ -396,12 +405,16 @@ module Phlex
 				name = case k
 					when String then k
 					when Symbol then k.name.tr("_", "-")
-					else raise ArgumentError, "Attribute keys should be Strings or Symbols."
+					else raise ArgumentError, "Attribute keys should be Strings, Symbols, or SafeValues."
 				end
 
 				# Detect unsafe attribute names. Attribute names are considered unsafe if they match an event attribute or include unsafe characters.
 				if HTML::EVENT_ATTRIBUTES[name] || name.match?(/[<>&"']/)
-					raise ArgumentError, "Unsafe attribute name detected: #{k}."
+					if v.is_a? Phlex::SGML::SafeValue
+						v = v.safe_value
+					else
+						raise ArgumentError, "Unsafe attribute name detected: #{k}."
+					end
 				end
 
 				case v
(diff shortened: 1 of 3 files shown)


bradgessler:
A few questions this PR raises.

Is `unsafe_raw` needed or should that be deprecated and require a value be passed into a `SafeValue` method.

I have a `safe` method, but is that too ambiguous? There's "safe HTML" and "safe attributes", do we need to distinguish between the two? I'd like to have a method that accepts a hash/keyword args of attributes and transforms all of their values as "safe". 

That might mean we end up with a few helpers that would appear in Phlex as follows:

```ruby
raw safe_html("...")
div(onclick: safe_attribute("alert()"))
div(**safe_attributes(onclick: "alert()", onblur: "alert()"))
```

lib/phlex/sgml.rb
```diff
@@ -396,12 +405,16 @@ def __build_attributes__(attributes, buffer:)
...
 				end
 
 				# Detect unsafe attribute names. Attribute names are considered unsafe if they match an event attribute or include unsafe characters.
 				if HTML::EVENT_ATTRIBUTES[name] || name.match?(/[<>&"']/)
-					raise ArgumentError, "Unsafe attribute name detected: #{k}."
+					if v.is_a? Phlex::SGML::SafeValue
+						v = v.safe_value
+					else
+						raise ArgumentError, "Unsafe attribute name detected: #{k}."
+					end
```
joeldrapper, line 417:
```suggestion
					case v
					when Phlex::SGML::SafeValue
						v = v.safe_value
					else
						raise ArgumentError, "Unsafe attribute name detected: #{k}."
					end
```

lib/phlex/sgml/safe_value.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# @api private
+class Phlex::SGML::SafeValue
```
joeldrapper, line 4: Can we move this up to `Phlex::SafeValue`?

lib/phlex/sgml.rb
```diff
@@ -396,12 +405,16 @@ def __build_attributes__(attributes, buffer:)
 				name = case k
 					when String then k
 					when Symbol then k.name.tr("_", "-")
-					else raise ArgumentError, "Attribute keys should be Strings or Symbols."
+					else raise ArgumentError, "Attribute keys should be Strings, Symbols, or SafeValues."
```
joeldrapper, line 408:
```suggestion
					else raise ArgumentError, "Attribute keys should be Strings, Symbols, or Phlex::SafeValues."
```

joeldrapper (commented):
This looks really good. Left a few comments.

lib/phlex/sgml/safe_value.rb
bradgessler: I put this in SGML thinking, "is a safe SGML value have the same characteristics as a safe CSV (or other format) value?"

I can move it if you thought of that and still think it should be `Phlex::SafeValue`

bradgessler:
Threw a point of consideration back at you. If you thought of it and are A-ok with it I'll update the PR and get this ready to merge.

lib/phlex/sgml/safe_value.rb
joeldrapper: Oh, you're probably right. I was thinking it's considered safe only because it's marked as safe, so doesn't matter which output language it's for. But this is probably better.

tcannonfodder:
Very excited to see this shipped! Just ran into this while working on as-error-tolerant-as-possible buttons to open a modal dialog, based on: https://codepen.io/tcannonfodder/pen/gONoedJ

joeldrapper:
I’m closing this as there have been quite a few changes to attributes since it was opened. Probably best to use this as a reference in a new take.

PR #729, opened 2024-06-09
Fix component kit autoload order bug

### LINKED ISSUE #727: Weird autoloading bug
Prerequisites:
we have a `app/views/components` directory which is added to the autoloaded paths during initialization of the app, i.e.:
```ruby
Rails.autoloaders.main.tap do |loader|
  loader.push_dir(Rails.root.join("app", "views").to_s)
end
```
there's also a `app/views/components.rb` which is just:
```ruby
module Components
  extend Phlex::Kit
end
```
now if we do something like this:
```ruby
# app/views/components/a.rb
class Components::A < Phlex::HTML
  def self.a
    # whatever
  end
end

# app/views/components/b.rb
class Components::B < Components::A
  a
end
```
we get an error saying `a is undefined`. Funnily enough I think it's order dependent if we switch `A` and `B`, i.e. make `B` inherit from `A`, everything works fine. I'm not sure why that happens and maybe there's something wrong on our side, but [here's](https://github.com/Vagab/phlex-wat) a minimal reproduction. It's a fresh `rails new` with the described above changes.
> Note: Oh, I've also added `eager_load = true` to the development just so you can see it crash as soon as you `rails c`. Without `eager_load` it will crash when irb loads it which is usually when you type `Components` into it.

joeldrapper:
Instead of loading all the constants, now we use method missing to lazy-load them.

I believe this fixes #727

### DIFF
--- a/lib/phlex/kit.rb
+++ b/lib/phlex/kit.rb
@@ -3,12 +3,23 @@
 module Phlex::Kit
 	def self.extended(mod)
 		warn "⚠️ [WARNING] Phlex::Kit is experimental and may be removed from future versions of Phlex."
-		super
-	end
 
-	# When a kit is included in a module, we need to load all of its components.
-	def included(mod)
-		constants.each { |c| const_get(c) if autoload?(c) }
+		mod.define_method(:respond_to_missing?) do |name, _include_private = false|
+			if name[0] == name[0].upcase && mod.constants.include?(name) && mod.const_get(name) && methods.include?(name)
+				true
+			else
+				super
+			end
+		end
+
+		mod.define_method(:method_missing) do |name, *args, **kwargs, &block|
+			if name[0] == name[0].upcase && mod.constants.include?(name) && mod.const_get(name) && methods.include?(name)
+				public_send(name, *args, **kwargs, &block)
+			else
+				super
+			end
+		end
+
 		super
 	end
 
--- a/lib/phlex/kit.rb
+++ b/lib/phlex/kit.rb
@@ -1,42 +1,39 @@
 # frozen_string_literal: true
 
 module Phlex::Kit
-	def self.extended(mod)
-		warn "⚠️ [WARNING] Phlex::Kit is experimental and may be removed from future versions of Phlex."
-
-		mod.define_method(:respond_to_missing?) do |name, _include_private = false|
-			if name[0] == name[0].upcase && mod.constants.include?(name) && mod.const_get(name) && methods.include?(name)
-				true
-			else
-				super
-			end
-		end
-
-		mod.define_method(:method_missing) do |name, *args, **kwargs, &block|
-			if name[0] == name[0].upcase && mod.constants.include?(name) && mod.const_get(name) && methods.include?(name)
+	module LazyLoader
+		def method_missing(name, *args, **kwargs, &block)
+			if name[0] == name[0].upcase && __phlex_kit_constants__.include?(name) && __get_phlex_kit_constant__(name) && methods.include?(name)
 				public_send(name, *args, **kwargs, &block)
 			else
 				super
 			end
 		end
 
-		super
-	end
-
-	def method_missing(name, *args, **kwargs, &block)
-		if name[0] == name[0].upcase && constants.include?(name) && const_get(name) && methods.include?(name)
-			public_send(name, *args, **kwargs, &block)
-		else
-			super
+		def respond_to_missing?(name, include_private = false)
+			if name[0] == name[0].upcase && __phlex_kit_constants__.include?(name) && __get_phlex_kit_constant__(name) && methods.include?(name)
+				true
+			else
+				super
+			end
 		end
 	end
 
-	def respond_to_missing?(name, include_private = false)
-		if name[0] == name[0].upcase && constants.include?(name) && const_get(name) && methods.include?(name)
-			true
-		else
-			super
-		end
+	include LazyLoader
+
+	def self.extended(mod)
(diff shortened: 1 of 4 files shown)


lib/phlex/kit.rb
```diff
@@ -1,31 +1,39 @@
...
+			end
+		end
+
+		def respond_to_missing?(name, include_private = false)
+			if name[0] == name[0].upcase && __phlex_kit_constants__.include?(name) && __get_phlex_kit_constant__(name) && methods.include?(name)
+				true
+			else
+				super
+			end
+		end
```
joeldrapper, line 19: This module hooks into `method_missing` and `respond_to_missing?` to lazy-load an *autoloaded* constant with the same name as the method.

The first check `name[0] == name[0].upcase` is for performance as it allows it to exist early in the case that the missing method doesn’t even start with a capital letter.

Next, it checks if a constant matching the method name exists. It goes through `__phlex_kit_constants__`, since this is sometimes called on a class and sometimes on an instance, and they both need to handle this call.

If a matching constant is found, it then gets the constant via `__get_phlex_kit_constant__`. This will trigger an autoloaded constant to be loaded. If this constant is indeed a `Phlex::SGML` component, loading it will automatically define the missing method that we’re responding to so it won’t be missing anymore.

Finally, we check if that method was defined with `methods.include?(name)`. If it was, we can call it again. It won’t cause an infinite loop because this time, we won’t end up in the `method_missing` function.

lib/phlex/kit.rb
```diff
@@ -1,31 +1,39 @@
...
-	def included(mod)
-		constants.each { |c| const_get(c) if autoload?(c) }
-		super
+	include LazyLoader
+
+	def self.extended(mod)
+		warn "⚠️ [WARNING] Phlex::Kit is experimental and may be removed from future versions of Phlex."
+		mod.include(LazyLoader)
+		mod.define_method(:__phlex_kit_constants__) { mod.__phlex_kit_constants__ }
+		mod.define_method(:__get_phlex_kit_constant__) { |name| mod.__get_phlex_kit_constant__(name) }
```
joeldrapper, line 28: We need to define these at extend time so the instance can delegate to the kit module.

lib/phlex/kit.rb
```diff
@@ -1,31 +1,39 @@
...
 
-	# When a kit is included in a module, we need to load all of its components.
-	def included(mod)
-		constants.each { |c| const_get(c) if autoload?(c) }
-		super
+	include LazyLoader
+
+	def self.extended(mod)
+		warn "⚠️ [WARNING] Phlex::Kit is experimental and may be removed from future versions of Phlex."
+		mod.include(LazyLoader)
```
joeldrapper, line 26: The reason to include this as a module rather than just define `method_missing` and `respond_to_missing?` directly like below is so it can call `super`.

lib/phlex/kit.rb
```diff
@@ -1,31 +1,39 @@
 # frozen_string_literal: true
 
 module Phlex::Kit
-	def self.extended(mod)
-		warn "⚠️ [WARNING] Phlex::Kit is experimental and may be removed from future versions of Phlex."
-		super
+	module LazyLoader
+		def method_missing(name, *args, **kwargs, &block)
+			if name[0] == name[0].upcase && __phlex_kit_constants__.include?(name) && __get_phlex_kit_constant__(name) && methods.include?(name)
```
willcosgrove, line 6: I spent way too long trying to performance golf `name[0] == name[0].upcase`.

I thought it was going to be easier than it was when I remembered that Ruby supports non-ascii capital letters as well. I did manage to figure out how to regex it, which does perform about twice as fast as this string transformation + comparison. But after all that, I think what you have is better, because it's clearer what is being done.

```rb
name.match?(/\A(\p{Lu}|\p{Lt})/) # \p{} checks for unicode properties, Lu is uppercase, Lt is titlecase

# Comparison:
#                regex: 13742991.5 i/s
#          start_with?:  7782039.9 i/s - 1.77x  slower
#    [0] == [0].upcase:  6729981.7 i/s - 2.04x  slower
```

willcosgrove (approved):
Wow, nice job figuring all this out. I ran through the original bug report, and I'm still not sure I understand why it was not working.

But what your change is doing makes sense. If I understand it correctly, this changes Phlex::Kit from loading all constants at the kit's load time, to loading the specific constants as their corresponding method gets called for the first time.

joeldrapper:
> I ran through the original bug report, and I'm still not sure I understand why it was not working.

Here’s the the code that triggered the bug:

```ruby
class Components::A < Phlex::HTML
  def self.a
    # whatever
  end
end

# app/views/components/b.rb
class Components::B < Components::A
  a
end
```

Here’s what I _think_ happened.

When `Components::B` was loaded, it autoloaded `Components::A` since it inherits form `Components::A`. When `Components::A` was being autoloaded, it included `Components` (this is part of the `const_added`. Including `Components` meant eagerly loading all the Components in order. `Components::A` was already loaded at this point but hadn’t yet defined `self.a`, so we hit `Components::B` a second time, this time without having to load `Components::A` so it continued to call `self.a`, which wasn’t yet defined.

lib/phlex/kit.rb
natematykiewicz: Isn't `name[0].casecmp?(name[0])` the fastest way to do a case-insensitive comparison?

natematykiewicz: Oh, I misunderstood what this was doing. Nevermind.

Issue #607, opened 2023-11-28
Disallow rendering the same component more than once


joeldrapper:
Rendering a component more than once can result in unintended consequences. For example, if your component uses `DeferredRender` to capture slots, the content of those slots will leak between different renders of the same instance.

I think we should update Phlex to disallow this, raising an error if you attempt to render the same instance of a component more than once. The place to do this would be in `__final_call__` which is the method that gets renamed to `call` after the first render for caching.

https://github.com/phlex-ruby/phlex/blob/6a0eb85d2fe6e3959644c45185150ee29856eee0/lib/phlex/sgml.rb#L97-L127

I would approach this by first checking for an instance variable and raising if it's true, and then setting that instance variable to true.

joeldrapper:
We can add this as a warning in 1.9.

bugdimes:
I will do this in a few days

PR #735, opened 2024-06-14
Warn on double render

Vagab:
wasn't sure if I should add tests for this. Adds a warning on double render to 1.10 #607

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -105,6 +105,11 @@ module Phlex
 			@_context = context
 			@_view_context = view_context
 			@_parent = parent
+			if @_rendered
+			  warn "⚠️ [WARNING] You are rendering a component #{self.class.name} twice. This is not supported in Phlex 2.0."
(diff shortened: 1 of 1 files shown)


lib/phlex/sgml.rb
```diff
@@ -105,6 +105,11 @@ def __final_call__(buffer = +"", context: Phlex::Context.new, view_context: nil,
 			@_context = context
 			@_view_context = view_context
 			@_parent = parent
+			if @_rendered
+			  warn "⚠️ [WARNING] You are rendering a component #{self.class.name} twice. This is not supported in Phlex 2.0."
+			  @_rendered = true
```
joeldrapper, line 110: I think this assignment was meant to be outside the `if` condition.

joeldrapper (changes requested):

lib/phlex/sgml.rb
Vagab: oops, right 😅

PR #736, opened 2024-06-20
Allow booleans as conditionals in the tokens helper

bjarkevad:
Often times I need to conditionally add classes based on a boolean value, and wrapping them in a Proc adds a bit of overhead. I hope it makes sense to allow raw booleans too.

I added a sus test to keep them in line with the other tests for `tokens`. Let me know if I should move it to quickdraw instead.

### DIFF
--- a/lib/phlex/helpers.rb
+++ b/lib/phlex/helpers.rb
@@ -20,10 +20,11 @@ module Phlex::Helpers
 	# 	)
 	def tokens(*tokens, **conditional_tokens)
 		conditional_tokens.each do |condition, token|
-			truthy = case condition
+  		truthy = case condition
 				when Symbol then send(condition)
 				when Proc then condition.call
-				else raise ArgumentError, "The class condition must be a Symbol or a Proc."
+				when TrueClass, FalseClass then condition
+				else raise ArgumentError, "The class condition must be a Symbol, Proc, TrueClass or FalseClass."
 			end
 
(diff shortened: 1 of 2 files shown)


joeldrapper:
Hey, thanks for the PR. Unfortunately because Hash keys need to be unique, this won’t work.

Let’s say I have the following variables:
```ruby
foo = true
bar = true
baz = false
bing = false
```

And I use them like this:
```ruby
div(
  **classes(
    foo => "foo",
    bar => "bar",
    baz => "baz",
    bing => "bing"
  )
)
```

Because keys are unique, the `classes` method will only receive the arguments:
```
(true => "bar", false => "bing")
```

So the output would be:
```html
<div class="bar"></div>
```

Instead of the expected:
```html
<div class="foo bar"></div>
```

Using Procs means each key is unique so it works as expected:

```ruby
div(
  **classes(
    -> { foo } => "foo",
    -> { bar } => "bar",
    -> { baz } => "baz",
    -> { bing } => "bing"
  )
)
```

Another way you could do this is with parentheses and regular conditionals:

```ruby
div(
  class: [
    ("foo" if foo),
    ("bar" if bar),
    ("baz" if baz),
    ("bing" if bing),
  ]
)
```

bjarkevad:
Ah of course, that was a bit of a brain fart. Thanks for taking the time to write out that massive response.

Issue #720, opened 2024-05-15
Method to make binding the `class` kwarg easier


bradgessler:
Its common to want to intercept the `class` kwarg like this:

```ruby
  def input(*, class: "hi", **, &)
    super(*, class: tokens(keyword(class:)))
  end
```

Phlex could have a method that makes this possible like this:

```ruby
    def keyword(**bindings)
      if bindings.size > 1
        bindings.values
      else
        bindings.values.first
      end
    end
```

Not sure if `keyword` is the right method name. Some possible method names are:

```
bind
keyword
keywords
value
values
grab
```

Previous attempt at https://github.com/phlex-ruby/phlex/pull/717 broke the way `tokens` behaved and was less composable. I believe this approach solves both of those problems.

joeldrapper:
I quite like this with the name `grab`. @willcosgrove what are your thoughts?

willcosgrove:
I like `grab` too. It is solving an interesting problem. I'm kind of surprised I haven't run into a need for it yet 🤔

In the example given, I would just use `mix` and not have any need to grab out the class kwarg.

joeldrapper:
Oh, of course you can already use `mix` for this. I’m trying to be very careful about what features we introduce in core, but I don’t think this represents a significant increase in complexity or maintenance burden. It’s the kind of thing that’s useful if you need it and disappears if you don’t, which I like.

joeldrapper:
Even though you can technically use `mix`, with `**mix(:class)`, I think something like this is quite a bit simpler when all you’re doing is passing along the `class`.

```ruby
def foo(class:)
  div(class: [grab(:class), "some other classes here"])
end
```

PR #737, opened 2024-06-23
feat: add Hash support for `class` and `style` attributes

ElMassimo:
### Description :book:

This pull request adds support for passing `Hash` in `class` and `style` attributes in a way that feels idiomatic in Ruby and achieves composability.

### Background 📜 

When writing Phlex components, things usually start simple:

```ruby
div(id: "hero", class: "section")
```

but when you need a conditional class, things get a bit awkward:

```ruby
div(id: "hero", **classes("section", ("active" if active)))
```

This proposal is meant to provide an alternative which is easier to understand, and easier to write:

```ruby
div(id: "hero", class: ["section", active:])
```

#### `class`

A component such as:

```ruby
def active = true

def inactive = false

def view_template
  div(class: ["dropdown", inactive:, active:])
end
```

would render:

```erb
<div class="dropdown active"></div>
```

#### `style`

A component such as:

```ruby
def view_template
  div(style: {font_size: "16px", opacity: 0})
end
```

would render:

```erb
<div style="font-size:16px;opacity:0;"></div>
```

### Precedents

This feels like the Ruby equivalent of what frameworks like [Vue](https://vuejs.org/guide/essentials/class-and-style.html) provide in their template system:

- [Class and Style Bindings](https://vuejs.org/guide/essentials/class-and-style.html)

As noted by @Spone, this behavior is consistent with the [`class_names`](https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-token_list) helper in Rails.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -520,7 +520,7 @@ module Phlex
 				end
 			end
 
-			style[-1] == ";" ? style : style + ";"
+			style.end_with?(";") ? style : style + ";"
 		end
 	end
 end
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -471,7 +471,7 @@ module Phlex
 			when nil, false
 				nil
 			else
-				buffer << if c.respond_to?(:to_phlex_attribute_value)
+				if c.respond_to?(:to_phlex_attribute_value)
 					c.to_phlex_attribute_value
 				elsif c.respond_to?(:to_str)
 					c.to_str
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -406,16 +406,32 @@ module Phlex
 				when Integer, Float
 					buffer << " " << name << '="' << v.to_s << '"'
 				when Hash
-					__attributes__(
-						v.transform_keys { |subkey|
-							case subkey
-								when Symbol then"#{name}-#{subkey.name.tr('_', '-')}"
-								else "#{name}-#{subkey}"
-							end
-						}, buffer
-					)
+			    case k
+					when :class
+						buffer << " " << name << '="' << __classes__(v).gsub('"', "&quot;") << '"'
+					when :style
+					  buffer << " " << name << '="' << __styles__(v).gsub('"', "&quot;") << '"'
+					else
+						__attributes__(
+							v.transform_keys { |subkey|
+								case subkey
+									when Symbol then"#{name}-#{subkey.name.tr('_', '-')}"
+									else "#{name}-#{subkey}"
+								end
+							}, buffer
+						)
+					end
 				when Array
-					buffer << " " << name << '="' << v.compact.join(" ").gsub('"', "&quot;") << '"'
+			    value = case k
+					when :class
+						__classes__(v)
+					when :style
+					  __styles__(v)
+					else
+						v.compact.join(" ")
+					end
+
+					buffer << " " << name << '="' << value.gsub('"', "&quot;") << '"'
 				when Set
 					buffer << " " << name << '="' << v.to_a.compact.join(" ").gsub('"', "&quot;") << '"'
 				else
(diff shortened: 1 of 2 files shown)


joeldrapper:
I really like the look of this!

lib/phlex/sgml.rb
```diff
@@ -433,5 +449,78 @@ def __attributes__(attributes, buffer = +"")
...
+				return nil
+			else
+				if s.respond_to?(:to_phlex_attribute_value)
+					s.to_phlex_attribute_value
+				elsif s.respond_to?(:to_str)
+					s.to_str
+				else
+					s.to_s
+				end
+			end
```
joeldrapper, line 521: This chunk is repeated 3x now. I wonder if it might be worth extracting a method for it, but that could impact performance.

joeldrapper (approved):
This looks good to me. I’ll hold off on merging until @willcosgrove has taken a look as he always seems to catch things I miss. 😅

I wonder if the browser tests (`browser_tests.rb`) should be updated to include some examples with class and style tags as arrays, hashes, symbols and strings so all the avenues for XSS are covered. There are a few points here where we push to the output buffer directly.

lib/phlex/sgml.rb
joeldrapper: To be clear, I’m fine with this repetition as long as it’s worth it for the perf benefits.

joeldrapper:
@bradgessler this makes your idea in #720 even more useful.

lib/phlex/sgml.rb
ElMassimo: I assumed inlining would benefit performance, but I agree we should benchmark it :+1:

The introduction of YJIT challenges some of the usual outcomes, perhaps extracting to a method no longer has a significant impact.

bradgessler:
This syntax has never really jived with my expectations—I think it's too clever.

For example, what does this return?

```ruby
def active = true

def inactive = false

def priority = %[low medium high].sample

def view_template
  active = false
  inactive = true
  div(class: ["dropdown", inactive:, active:, priority:])
end
```

I find it surprising that it binds to the value of a method and shows the class, or not, if its true or false.

Instead I'd propose methods that make it clearer what these things are doing. Here's an idea if method is used:

```ruby
def active = true

def inactive = false

def priority = %[low medium high].sample

def view_template
  # This would inflect on the Method class. Token is Method#name and if the return
  # value is a string, it would populate with that string. If it's boolean it would return
  # Method#name token if it's true. Otherwise nothing if its empty.
  div(class: ["dropdown", method(:active), method(:inactive), method(:priority))]
end
```

If I'm reading this code it's much clearer to me, "Oh, this is the `:active` method". I can even copy and paste the code into my debugger to look at it.

There could probably be some Phlex helper methods that dress this up a bit more, but I think at a minimum using the already existing `method` method would improve readability and prevent this API from being painted into a corner.

hangsu:
+1 for `style` Hash support. We've used a custom implementation at Clearscope for some time now.

I tend to agree with @bradgessler's sentiments around `class`. We've always used arrays of strings and have never been tempted to do more:
```ruby
def view_template
  div(class: ["btn", ("active" if active?)])
end

def active? = true

# OR

def view_template
  div(class: ["btn", active_class])
end

def active_class = "active" if condition?
```

joeldrapper:
@bradgessler I think the example wasn’t clear because it used the same class name as conditional. It’s not calling the key as a method. I think this might be a better example:

```
div(class: ["rounded", "bg-blue color-white font-bold" => @active])
```

Read `=>` above as “if”.

Spone:
Quick note: this is similar to the [`token_list`](https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-token_list) Rails helper (aliased `class_names`) which is in my experience quite easy to work with.

lib/phlex/sgml.rb
```diff
@@ -433,5 +449,78 @@ def __attributes__(attributes, buffer = +"")
...
+
+			style[-1] == ";" ? style : style + ";"
```
willcosgrove, line 523: Two things here:

1. Is it important that style strings end with a semicolon? It looks like they work either way.
2. We can save an allocation and get a speed up by changing this check to use `end_with?` instead of comparing against a slice of the string.

lib/phlex/sgml.rb
willcosgrove: I'm also curious to know if extracting it does help or not. I could see that that would make it a more frequented code path so it might get JITed sooner. I don't think we need to hold this up for that experiment though.

willcosgrove (approved):
I really like this. I've been hesitant in the past to suggest this change because I didn't want to over complicate the mental rules for how ruby translates to HTML. But I think the developer experience is worth the break from the normal rules for these two heavily used HTML attributes.

I'm looking forward to using it!

lib/phlex/sgml.rb
ElMassimo: 1. When combining a user-provided string in the Array and Hash variants, joining with `;` might result in `;;` if the user already had that suffix. While that's still valid CSS, it felt nicer for `__style__` output to be consistent
2. 👍 on using `end_with?` if it's faster

bradgessler:
> @bradgessler I think the example wasn’t clear because it used the same class name as conditional. It’s not calling the key as a method. I think this might be a better example:
> 
> ```
> div(class: ["rounded", "bg-blue color-white font-bold" => @active])
> ```
> 
> Read `=>` above as “if”.

Oh I see, so the "magic method" stuff is gone since the `tokens` helper is removed? What you propose is much more sane and less surprising since the behavior is congruent to how it works for HTML attributes.

PR #738, opened 2024-06-24
Remove `tokens` and `classes` helpers

joeldrapper:
Since #737, we no longer need these helpers.


joeldrapper:
@willcosgrove can you think of any reason why we should keep these?

willcosgrove:
I can't think of a long term reason to keep them around. It might be nice if they didn't disappear at the same time as a new way gets introduced, just so the migration isn't as cumbersome

joeldrapper:
> It might be nice if they didn't disappear at the same time as a new way gets introduced, just so the migration isn't as cumbersome

That’s true, but introducing the new way and removing the old way are both breaking changes. We’d either need to keep them around until 3.0 or remove them with 2.0 or sneak the new way out in a 1.x release even though it’s technically breaking.

willcosgrove:
>passing a Hash was previously resulting in to_s to be called
I think it would have made hyphenated HTML attributes:

```rb
div(class: { foo: true }) # => <div class-foo></div>
```

But I think your point still stands: while technically breaking existing behavior, it is unlikely anyone was relying on it for `class` and `style` specifically.

willcosgrove:
Can we make them issue a deprecation warning and remove them in a later point release? Or maybe I'm the only one using them 😳 In which case, I don't need a slow removal of them for myself.

brandondrew:
> > It might be nice if they didn't disappear at the same time as a new way gets introduced, just so the migration isn't as cumbersome
> 
> That’s true, but introducing the new way and removing the old way are both breaking changes. We’d either need to keep them around until 3.0 or sneak the new way out in a 1.x release even though it’s technically breaking.

Does having the new way available really break any existing code?

joeldrapper:
> Does having the new way available really break any existing code?

Technically/theoretically, yes. Realistically, no way.

Let’s port the new behaviour back to 1.x and add a deprecation warning to these helpers. I’m happy to work on that. We’re due a 1.11 release anyway.

joeldrapper:
I’m going to ship this to `main`, which is tracking `2.0`. But I’ll back port the other changes and deprecations.

PR #742, opened 2024-07-04
CI: Use stable version of Quickdraw

bquorning:
But, also add a separate job that runs with edge version of Quickdraw.
Some tests were failing because they had no expectation (I added one), and some were failing because they didn’t match the implementation (I changed the tests to match).


bquorning:
Finding a stable version of Quickdraw _and_ fixing the Phlex tests was a bit harder than anticipated. But it seems to work now (apart from the truffleruby-23.1 error which I haven’t looked into).

quickdraw/helpers/grab.test.rb
```diff
@@ -3,11 +3,11 @@
...
-	output = grab({ class: "foo" }, { if: "bar" })
+	output = grab(class: "foo", if: "bar")
 	expect(output) == ["foo", "bar"]
 end
 
 test "supports single binding" do
-	output = grab({ class: "foo" })
-	expect(output) == ["foo"]
+	output = grab(class: "foo")
+	expect(output) == "foo"
```
bquorning, line 12: I am unsure about this change. Is it intentional that `#grab` sometimes returns an Array and sometimes a String?

bquorning:
I will be unable to work on this PR for the next week or so. @joeldrapper If you feel like it, please take over and change it as you like.

joeldrapper:
Thank you!

PR #746, opened 2024-08-11
Add `__nested__attributes__`

AnthonySuper:
Special-cases the serialization of nested attributes to reduce the number of allocations. This avoids having to build up a new hash on each recursive call, and lets us "bypass" a lot of the safety features (such as dangerous attribute sanitation) that Phlex includes when serializing nested attributes on elements (since, in the nested case, we *know* that our key cannot be something like `href` or `onclick`). 

This should reduce the number of allocations and make the code more performant.

PR #747, opened 2024-08-12
Optimize nested tokens

joeldrapper:
After reading through #746 I had an idea about how to further optimise nested tokens (Arrays and Sets).

1. `.compact.join` was doing two iterations over the array: one to remove nils and one to build a string.
2. `.gsub` was creating a new string unnecessarily.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -501,8 +501,6 @@ module Phlex
 				token = tokens[i]
 
 				case token
-				when nil
-					next
 				when String
 					if i > 0
 						buffer << " " << token
@@ -515,6 +513,7 @@ module Phlex
 					else
 						buffer << token.name
 					end
+				when nil
 				else
 					if i > 0
 						buffer << " " << token.to_s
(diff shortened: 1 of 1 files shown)


lib/phlex/sgml.rb
```diff
@@ -491,6 +491,45 @@ def __nested_attributes__(attributes, base_name, buffer = +"")
...
+						buffer << " " << token
+					else
+						buffer << token
+					end
+				when Symbol
+					if i > 0
+						buffer << " " << token.name
+					else
+						buffer << token.name
+					end
```
willcosgrove, line 517: I spent too long in a `bench.rb` file trying to beat this. But I think you've already wrung out all the perf that can be gained here. I'm frustrated that all these `if i > 0` checks are faster than always including the space and using `delete_suffix!(" ")` or `delete_prefix!(" ")` (depending on where you put the space).

That being said, I think for _most_ token lists, moving the `nil` check to the end will be faster. Also there is a bug where `i` never gets incremented when you encounter a nil.

```suggestion
				when String
					if i > 0
						buffer << " " << token
					else
						buffer << token
					end
				when Symbol
					if i > 0
						buffer << " " << token.name
					else
						buffer << token.name
					end
				when nil
```

willcosgrove (approved):

PR #748, opened 2024-08-12
Add dynamic tag method

### LINKED ISSUE #707: Add a safe dynamic element method
It was brought up in the Discord that we should have a safe way to create dynamic elements. So instead of doing:
```rb
send(tag_name, **attributes)
```
which could potentially call _any_ method, not just an element method, we could have something more like:
```rb
element(tag_name, **attributes)
```
Where `tag_name` could be a string or symbol that matches any registered element, and it would raise if it didn't match an element.

The name is up for debate. Some ideas:

* `element` - nice and short, no current conflicts with HTML or SVG, but because it's not a compound name, it could possibly get used one day
* `safe_element` - compound names gets us out of conflict territory with HTML or SVG spec
* `dynamic_element`
* `safe_tag`
* `dynamic_tag`
* `tag` - shortest, but will conflict with the rails helper... maybe this is a good conflict if we can shim the API in `phlex-rails` to match the helper (The main inconsistency is that the Rails `tag` helper can take content as the first positional argument) 
  ```rb
  # phlex-rails override
  def tag(content=nil, **attributes, &block)
    raise ArgumentError, "Cannot pass content positionally and in a block" if content && block_given?

    if content
      super(**attributes) { content }
    else
      super(**attributes, &block)
    end
  end
  ```

joeldrapper:
Closes #707 

This provides a safer alternative to using `send` and `public_send` to dynamically call Phlex methods when using your own meta-programming.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -206,6 +206,38 @@ module Phlex
 			end
 		end
 
+		def tag(name, ...)
+			normalized_name = case name
+				when Symbol then name.name.downcase
+				when String then name.downcase
+				else raise ArgumentError.new("Expected the tag namea as a Symbol or String.")
+			end
+
+			if name == "script"
+				raise ArgumentError.new("You can’t use the `<script>` tag from the `tag` method. Use `unsafe_tag` instead, but be careful if using user input.")
+			end
+
+			if registered_elements[normalized_name]
+				public_send(normalized_name, ...)
+			else
+				raise ArgumentError.new("Unknown tag: #{normalized_name}")
+			end
+		end
+
+		def unsafe_tag(name, ...)
+			normalized_name = case name
+				when Symbol then name.name.downcase
+				when String then name.downcase
+				else raise ArgumentError.new("Expected the tag namea as a Symbol or String.")
+			end
+
+			if registered_elements[normalized_name]
+				public_send(normalized_name, ...)
+			else
+				raise ArgumentError.new("Unknown tag: #{normalized_name}")
+			end
+		end
+
 		private
 
 		# @api private
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -228,7 +228,7 @@ module Phlex
 			normalized_name = case name
 				when Symbol then name.name.downcase
 				when String then name.downcase
-				else raise ArgumentError.new("Expected the tag namea as a Symbol or String.")
+				else raise ArgumentError.new("Expected the tag name as a Symbol or String.")
 			end
 
 			if registered_elements[normalized_name]
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -213,7 +213,7 @@ module Phlex
 				else raise ArgumentError.new("Expected the tag namea as a Symbol or String.")
 			end
 
-			if name == "script"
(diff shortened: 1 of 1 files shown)


lib/phlex/sgml.rb
```diff
@@ -206,6 +206,38 @@ def capture(*args, &block)
 			end
 		end
 
+		def tag(name, ...)
+			normalized_name = case name
+				when Symbol then name.name.downcase
+				when String then name.downcase
```
fractaledmind, line 212:
```suggestion
				when Symbol then name.name.downcase!
				when String then name.downcase!
```

lib/phlex/sgml.rb
```diff
@@ -206,6 +206,38 @@ def capture(*args, &block)
...
+				else raise ArgumentError.new("Expected the tag namea as a Symbol or String.")
+			end
+
+			if name == "script"
```
fractaledmind, line 216:
```suggestion
			if normalized_name == "script"
```
Otherwise you can slip this check with "SCRIPT"

lib/phlex/sgml.rb
```diff
@@ -206,6 +206,38 @@ def capture(*args, &block)
...
+				public_send(normalized_name, ...)
+			else
+				raise ArgumentError.new("Unknown tag: #{normalized_name}")
+			end
+		end
+
+		def unsafe_tag(name, ...)
+			normalized_name = case name
+				when Symbol then name.name.downcase
+				when String then name.downcase
```
fractaledmind, line 230:
```suggestion
				when Symbol then name.name.downcase!
				when String then name.downcase!
```

lib/phlex/sgml.rb
joeldrapper: Ops! I renamed `name` to `normalized_name` and missed this one.

lib/phlex/sgml.rb
willcosgrove: `downcase!` also returns `nil` if no changes were made

lib/phlex/sgml.rb
```diff
@@ -206,6 +206,38 @@ def capture(*args, &block)
 			end
 		end
 
+		def tag(name, ...)
+			normalized_name = case name
+				when Symbol then name.name.downcase
+				when String then name.downcase
+				else raise ArgumentError.new("Expected the tag namea as a Symbol or String.")
```
willcosgrove, line 213:
```suggestion
				else raise ArgumentError.new("Expected the tag name as a Symbol or String.")
```

lib/phlex/sgml.rb
```diff
@@ -206,6 +206,38 @@ def capture(*args, &block)
...
+			else
+				raise ArgumentError.new("Unknown tag: #{normalized_name}")
+			end
+		end
+
+		def unsafe_tag(name, ...)
+			normalized_name = case name
+				when Symbol then name.name.downcase
+				when String then name.downcase
+				else raise ArgumentError.new("Expected the tag namea as a Symbol or String.")
```
willcosgrove, line 231:
```suggestion
				else raise ArgumentError.new("Expected the tag name as a Symbol or String.")
```

willcosgrove (approved):
This looks great! There's one typo in two of the error messages which I left comments on.

Issue #750, opened 2024-08-20
Inline tags declaration


Xanderwot:
Hi everyone,

I have a question (more like an advice) about phlex tag registration improvement.

We are doing a migration from another rendering gem ([fortitude](https://github.com/ageweke/fortitude)) to phlex and have a lot of legacy code, which supports both block and inline tags:
```ruby
div do
  'Some content'
end

# returns same result
div 'Some content'
```

I did patch `register_element` method [here](https://github.com/phlex-ruby/phlex/blob/1.11.0/lib/phlex/elements.rb#L31) like this:
```ruby
def #{method_name}(inline_content = nil, **attributes, &block)
#...
          elsif inline_content.present? # with inline content
            buffer << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[respond_to?(:process_attributes) ? (attributes.hash + self.class.hash) : attributes.hash] || __attributes__(**attributes)) << ">"
            inline_content_block = Proc.new { inline_content }
            yield_content(&inline_content_block)
            buffer << "</#{tag}>"
          else
#...
end
```
(without attributes part also updated)

 to support both approaches and its works pretty good in most of places except some specific cases.

After applying this patch to a Phlex::HTML::StandardElements
```ruby
Phlex::HTML::StandardElements.module_eval { extend PhlexElementRegisterer }
```

and re-registering all elements with patched register_element method:
```ruby
Phlex::HTML::StandardElements.registered_elements.each do |method_name, tag|
  Phlex::HTML::StandardElements.register_element(method_name, tag: tag)
end
```

For example nested tags like tables:
```ruby
table do 
  tr do 
    td 'Content'
  end
end
```

generate expected html:
```html
<table>
  <tr>
    <td>Content</td>
  </tr>
</table>
```

but if I do something more specific:
```ruby
table do 
  tr do 
    td div { 'Content' }
  end
end
```

it moves div block outside table tag:
```html
<div>Content</div>
<table>
  <tr>
    <td></td>
  </tr>
</table>
```

When debugging I see that the div tag method was called with &block and after it the td method also called, but without any arguments or even a block.

As a second example it happens not only with table tag:
```ruby
def link_tag(text)
  div do
    link_to(text, '#')
  end
end

div class: 'first_nested' do
  div class: 'second_nested' do
    div link_tag('something'), class: 'third_nested'
  end
end
```

Will return:
```html
<div class="first_nested">
  <div class="second_nested">
[...]

PR #755, opened 2024-08-30
Support inline tags

Xanderwot:
Based on [issue](https://github.com/phlex-ruby/phlex/issues/750)

Adding support of inline tag declaration
Examples:
```
h1 'Hello'
```
will render:
```
<h1>Hello</h1>
```

```
div do
  h1 'Hello', class: 'title'
end
```
will render:
```
<div><h1 class="title">Hello</h1></div>
```

### DIFF
--- a/lib/phlex/elements.rb
+++ b/lib/phlex/elements.rb
@@ -61,8 +61,7 @@ module Phlex::Elements
 						buffer << "</#{tag}>"
 					elsif content
 						buffer << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[attributes] ||= __attributes__(attributes)) << ">"
-						inline_block = content.is_a?(Proc) ? content : proc { content }
-						yield_content(&inline_block)
+						process_inline_content(content)
 						buffer << "</#{tag}>"
 					else # without content block
 						buffer << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[attributes] ||= __attributes__(attributes)) << "></#{tag}>"
@@ -74,8 +73,7 @@ module Phlex::Elements
 						buffer << "</#{tag}>"
 					elsif content
 						buffer << "<#{tag}>"
-						inline_block = content.is_a?(Proc) ? content : proc { content }
-						yield_content(&inline_block)
+						process_inline_content(content)
 						buffer << "</#{tag}>"
 					else # without content block
 						buffer << "<#{tag}></#{tag}>"
--- a/lib/phlex/elements.rb
+++ b/lib/phlex/elements.rb
@@ -32,7 +32,7 @@ module Phlex::Elements
 		class_eval(<<-RUBY, __FILE__, __LINE__ + 1)
 			# frozen_string_literal: true
 
-			def #{method_name}(**attributes, &block)
+			def #{method_name}(content = nil, **attributes, &block)
 				context = @_context
 				buffer = context.buffer
 				fragment = context.fragments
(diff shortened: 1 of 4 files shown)


lib/phlex/elements.rb
```diff
@@ -59,6 +59,11 @@ def #{method_name}(**attributes, &block)
 						buffer << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[attributes] ||= __attributes__(attributes)) << ">"
 						yield_content(&block)
 						buffer << "</#{tag}>"
+					elsif content
+						buffer << "<#{tag}" << (Phlex::ATTRIBUTE_CACHE[attributes] ||= __attributes__(attributes)) << ">"
+						inline_block = content.is_a?(Proc) ? content : proc { content }
+						yield_content(&inline_block)
```
joeldrapper, line 65: I think we should probably do something more like
```ruby
case content
when Block
  yield_content(&content)
when String
  plain(content)
else
  raise ArgumentError
end
```

Xanderwot: Good catch. 
Updated. 
There is no `Block` class in ruby. all `proc` and `lambda` is a `Proc`

I moved processing method to `SGML` class. I hope it's ok. 
Tests also improved

Xanderwot: Pls let me know if I need to update documentation somewhere. Thanks

joeldrapper (approved):
I like this change and it’s non-breaking. We’ve never come up with anything else to use the positional argument for so I think this makes sense. I’ve pinged @willcosgrove for his thoughts though as this is a huge change and one that would be difficult to reverse.

willcosgrove:
@joeldrapper convinced me that this is a good change 😄

I think we should not allow nesting using this though.

```rb
h1(em("Hello")) # 💥
```

And I think wrapping it in a proc (while it does fix the issue of the elements not getting added to the buffer at the right time) is not something we should support.

Xanderwot:
> @joeldrapper convinced me that this is a good change 😄
> 
> I think we should not allow nesting using this though.
> 
> ```ruby
> h1(em("Hello")) # 💥
> ```
> 
> And I think wrapping it in a proc (while it does fix the issue of the elements not getting added to the buffer at the right time) is not something we should support.

It's actually pretty specific cases what I faced when tests inline tags in my application. I doubt what it frequently used(`h1(em("Hello"))` I mean). Anyway proc wrapping is not a main idea. Idea of inline tags to make page structure little bit easier in some cases and not to use it in some crazy cases like huge inline tags sequence.
I did include this specific to a tests just to show how it can be solved because of ruby specific but using it delegates to a developers(if they know what they doing)
It's yours decision anyway I'm just suggesting small improvement to you(I hope so) maybe it will be useful for someone. 
Actually it comes from our migration from `fortitude` which supports this inline tags

joeldrapper:
Okay final feedback on this:

Let's make sure we raise if you use a positional argument **_and_** a block.

```ruby
h1("Hello") { "World" }
```

Let’s make sure we raise if you pass `nil`.

```ruby
h1(h2("Hi"))
```

`h2` here returns `nil` so the `h1` call should raise. The way we can detect this is by setting the default to a different null that we can specifically check for.

Note: in order to do this, we won’t be able to use `nil` as the default (`content = nil`). Instead, let’s define a `Phlex::Null` object and use that as the default.

```ruby
module Phlex
  Null = Object.new.freeze
end
```

Then we can check against that

```
  if Phlex::Null != content
    ...
```

Let’s also remove the `Proc` option as @willcosgrove mentioned at least for now. Adding support for this later is non-breaking but we can’t remove it once we add it.

Finally, while extracting methods is generally good practice, I expect we will get better performance if we just inline the stuff in `process_inline_content` even though there will be some duplication between standard and void elements. This will save a method call for every time we render an element with positional content.

Xanderwot:
@joeldrapper Pls take a look ;)

PR #757, opened 2024-08-30
Register SVG elements G | TEXT | CIRCLE

pierry01:
Recently I've created a custom Component on my APP and I needed to add a custom code, so here is my contribution

image

```
module Custom
  class Meter < ApplicationComponent
    extend Phlex::Elements

    register_element :g
    register_element :text
    register_element :circle

    def initialize(size:, text:, percentage:)
      super()

      @size = size
      @text = text
      @percentage = percentage
    end

    def view_template
      r = @size / 2
      cx_cy = r * 1.5
      svg_size = @size * 1.5
      stroke_dasharray = 2 * Math::PI * r
      stroke_percentage = ((100 - @percentage) * stroke_dasharray) / 100

      svg(width: svg_size, height: svg_size, style: "flex: none; margin: -#{svg_size / 8}") do
        g(transform: "rotate(-90 #{cx_cy} #{cx_cy})") do
          circle(
            r:,
            cx: cx_cy,
            cy: cx_cy,
            fill: "transparent",
            "stroke-width": r * 0.2,
            "stroke-dashoffset": "0",
            stroke: "rgb(226 232 240)",
            "stroke-dasharray": stroke_dasharray
          )

          circle(
            r:,
            cx: cx_cy,
            cy: cx_cy,
            fill: "transparent",
            "stroke-width": r * 0.2,
            stroke: "rgb(34 197 94)",
            "stroke-dasharray": stroke_dasharray,
            "stroke-dashoffset": stroke_percentage
          )
        end

        if @text.present?
          text(
            x: "50%",
            y: "50%",
            "font-size": @size / 3,
            "text-anchor": "middle",
            "dominant-baseline": "central"
          ) { @text }
        end
      end
    end
  end
end
```

### DIFF
--- a/lib/phlex/html/standard_elements.rb
+++ b/lib/phlex/html/standard_elements.rb
@@ -724,4 +724,25 @@ module Phlex::HTML::StandardElements
 	# 	@yieldparam component [self]
 	# 	@see https://developer.mozilla.org/docs/Web/HTML/Element/wbr
 	register_element :wbr
+
+	# @!method g(**attributes, &content)
+	# 	Outputs a `<g>` tag.
+	# 	@return [nil]
+	# 	@yieldparam component [self]
+	# 	@see https://developer.mozilla.org/en-US/docs/Web/SVG/Element/g
+	register_element :g
+
+	# @!method text(**attributes, &content)
+	# 	Outputs a `<text>` tag.
+	# 	@return [nil]
+	# 	@yieldparam component [self]
+	# 	@see https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text
+	register_element :text
+
+	# @!method circle(**attributes, &content)
+	# 	Outputs a `<circle>` tag.
+	# 	@return [nil]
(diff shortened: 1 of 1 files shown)


joeldrapper:
Hey @pierry01

I don’t think it's documented very well (if at all), but we actually do have support for SVG elements already, though they’re implemented in a slightly different way.

First, you can define an SVG component by inheriting from `Phlex::SVG` instead of `Phlex::HTML`. This works exactly the same as `Phlex::HTML`.

Also, if you use the `svg` tag inside a `Phlex::HTML` template, you are yielded a `Phlex::SVG` component that you can use inline.

```ruby
svg do |s|
  s.path(...)
  s.circle(...)
end
```

The reason we decided not to add all the SVG elements to `Phlex::HTML` directly is because they’re only valid inside an `<svg>` tag.

joeldrapper:
I’m going to close this for now. Feel free to open a new issue if you feel further discussion is needed about adding SVG tags to `Phlex::HTML` directly.

PR #763, opened 2024-09-04
Simplify FIFO cache

joeldrapper:
Use the attributes hash as the cache key directly.

**Before**
```
ruby 3.3.1 (2024-04-23 revision c56cd86388) +YJIT [arm64-darwin23]
                Page   335.000 i/100ms
                Page      3.369k (± 0.6%) i/s -     17.085k in   5.071279s
```

**After**
```
ruby 3.3.1 (2024-04-23 revision c56cd86388) +YJIT [arm64-darwin23]
                Page   275.000 i/100ms
                Page      2.760k (± 0.7%) i/s -     14.025k in   5.082058s
```

### DIFF
--- a/lib/phlex/fifo.rb
+++ b/lib/phlex/fifo.rb
@@ -18,8 +18,7 @@ class Phlex::FIFO
 	end
 
 	def [](key)
-		k, v = @hash[key.hash]
-		v if k == key
+		@hash[key]
 	end
 
 	def []=(key, value)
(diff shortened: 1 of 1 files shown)


joeldrapper:
Based on this, it’s ~22% faster overall to key off the digest, directly storing and comparing to the original ourselves rather than letting Ruby do that by keying off the attributes hash.

joeldrapper:
Looks like we might be able to do this once this ships in... 3.4 I guess? https://github.com/byroot/ruby/commit/67efd7d035ad1110c4aa575f891eb9a2152a0bd6

joeldrapper:
Closing this for now.

PR #765, opened 2024-09-04
Remove inline content

joeldrapper:
I’m really sorry @Xanderwot. We changed our minds about this.

There was some discussion about this feature here https://github.com/orgs/phlex-ruby/discussions/761. We also discovered it added a 16% performance overhead.

I really appreciate your contributions and I’m sorry this one didn’t make it.


Xanderwot:
👍 
I totally understand what its pretty specific and probably rare used feature
Anyway just as discussion possibility:
What do you think about moving this feature under configuration option(like `Phlex.config.enable_inline_tags`)? 
So it will allow to keep performance and make this feature optional. 

Thanks for cooperation anyway

lib/phlex/elements.rb
```diff
@@ -32,21 +32,13 @@ def register_element(method_name, tag: method_name.name.tr("_", "-"))
 		class_eval(<<-RUBY, __FILE__, __LINE__ + 1)
 			# frozen_string_literal: true
 
-			def #{method_name}(content = Phlex::Null, **attributes)
```
Xanderwot, line 35: You need to remove `Phlex::Null` definition also :)

joeldrapper:
> What do you think about moving this feature under configuration option(like Phlex.config.enable_inline_tags)?

The problem is any kind of check we do in those generated methods can have such a significant impact on the overall performance because they’re hit so many times.

I think there's a way you could patch it though so inline content still works.

As a little proof of concept, you should be able to do something like this

```ruby
def div(content = nil, *, **)
  if content
    super(*, **) { plain(content) }
  else
    super
  end
end
```

You could iterate over `Phlex::HTML.registered_elements` and redefine each one.

joeldrapper:
I guess we could break up the code generation into small parts that each can be patched. I’m trying to avoid over-complicating things though as that file is already pretty messy.

PR #766, opened 2024-09-05
Replace `_` with `-` in Symbol attribute values

joeldrapper:
This provides better consistency with Symbol keys and makes it easier to work with standard multi-word HTML attribute values, such as `contenteditable: :plaintext_only`.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -442,7 +442,7 @@ class Phlex::SGML
 			when String
 				buffer << " " << name << '="' << v.gsub('"', "&quot;") << '"'
 			when Symbol
-				buffer << " " << name << '="' << v.name.gsub('"', "&quot;") << '"'
+				buffer << " " << name << '="' << v.name.tr("_", "-").gsub('"', "&quot;") << '"'
 			when Integer, Float
 				buffer << " " << name << '="' << v.to_s << '"'
 			when Hash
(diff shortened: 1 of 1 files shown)


willcosgrove (approved):
Looks great! I don't think I've ever run into this, but I think I would have expected it to work the way that this PR changes it to work. 🎉

joelmoss:
Am I missing something, but what happened to the `process_attributes` hook?

willcosgrove:
@joelmoss it was renamed: https://github.com/phlex-ruby/phlex/pull/653#discussion_r1493843477

joeldrapper:
The `process_attributes` hook was removed. You can achieve the same by patching `__attributes__` but this is a private API and may change in the future.

joelmoss:
I missed that, but when and why was it removed? I rely on that and don't like the idea of monkey patching (`__attributes__`) private API :(

joeldrapper:
FWIW, it’s very unlikely we’ll rename `__attributes__` in the future, and it’s very likely to continue to be sent a single hash argument, which you can format before passing up to `super`.

joelmoss:
Can we at least make `__attributes__` public then please?

joeldrapper:
Let’s open an issue for that. There are so many caveats to patching that method, I’m not sure if it’s something we want to encourage.

joelmoss:
> @joelmoss it was renamed: [#653 (comment)](https://github.com/phlex-ruby/phlex/pull/653#discussion_r1493843477)

I even commented on it! 🤦‍♂️

PR #772, opened 2024-09-06
Safe values

### LINKED ISSUE #718: Escape hatch to bypass Phlex HTML safety
@bradgessler recently brought up the need for this again, and suggested having a `safe` method that we can use to wrap any safe value.

```ruby
div(onclick: safe("foo")) { "Foo" }
a(href: safe("javascript:bar")) { "Bar" }
```

I think this would work quite nicely. `safe` could return a `Phlex::SafeValue`, e.g.

```ruby
def safe(value)
  Phlex::SafeValue.new(value)
end
```

And we can check for that when building attributes.

```ruby
if Phlex::SafeValue === value
  ...
end
```

joeldrapper:
- Remove `Phlex::SGML#unsafe_raw`
- Add `Phlex::SGML#safe` which creates a `Phlex::SGML::SafeValue`
- Add `Phlex::SGML::SafeObject` a generic interface for safe objects, which `Phlex::SGML::SafeValue` implements
- Add `Phlex::SGML#raw`, which takes a SafeObject
- Update `__attributes__` to allow the use of unsafe attribute names when the values are SafeObjects.

Example
```ruby
def view_template
  raw 🦺 "<script>"
  a href: 🦺 "javascript:alert(1)"
  a onclick: 🦺 "alert(1)"
end
```

Thanks @bradgessler for writing the original implementation of this PR. #719 

Closes #718


willcosgrove (approved):
Looks great!

lib/phlex/sgml.rb
```diff
@@ -229,20 +237,12 @@ def tag(name, ...)
...
-		if registered_elements[normalized_name]
-			public_send(normalized_name, ...)
-		else
-			raise Phlex::ArgumentError.new("Unknown tag: #{normalized_name}")
-		end
+	def safe(value)
+		Phlex::SGML::SafeValue.new(value)
 	end
 
+	alias_method :🦺, :safe
```
joeldrapper, line 244: 🕶️

test/phlex/view/raw.rb
```diff
@@ -18,7 +30,7 @@ def view_template
 	with "nil content" do
 		view do
 			def view_template
-				unsafe_raw nil
+				raw nil
```
bradgessler, line 33: This method name is so much better than `unsafe_raw`!

davidalejandroaguilar:
Hey, good timing! Any idea when this will be released? Found myself in need for it for implementing the Modal component in PhlexyUI:

```html
<button class="btn" onclick="my_modal_1.showModal()">open modal</button>
<dialog id="my_modal_1" class="modal">
...
```

willcosgrove:
It's in the betas that are out for 2.0. I'm not sure there's a target date for releasing 2.0, other than just "when it's ready" 😅

Probably before ruby 3.4 comes out

davidalejandroaguilar:
Have to ask, would it be possible to add this pre-2.0?

bradgessler:
Install 2.x beta. There's nothing super risky in there; just a few changes that break a few things, but should be relatively straightforward to fix.

davidalejandroaguilar:
@bradgessler Yeah I also think there's not too many breaking changes. 

The thing is that I'm needing this for https://github.com/PhlexyUI/phlexy_ui, and even though it's new and probably no-one uses it yet, if someone does, they may not want to be in Phlex 2.x beta 😅

joeldrapper:
You can get around it in v1 by using `unsafe_raw`.

davidalejandroaguilar:
Thanks for chiming in @joeldrapper!

I did see that helper being removed in this PR, but had trouble getting it to work with yielded content. I'm opening a discussion here [Discussion #793: unsafe_raw with yield renders the yielded content outside](https://github.com/orgs/phlex-ruby/discussions/793) to avoid hijacking this PR further.

Thanks again everyone 😃

bradgessler:
Unrelated: @davidalejandroaguilar when you take on forms in PhlexyUI, take a look at https://github.com/rubymonolith/superform

I'm in the same boat as you--holding out for 2.x to drop before switching everything over.

davidalejandroaguilar:
@bradgessler I've seen it of course (I have it starred 😉), it's very interesting! Great to see so many different takes built on Phlex.

PR #773, opened 2024-09-06
Allow elements to output safe return values

willcosgrove:
### Before

```ruby
script { safe(%(console.log("Hi 👋"))) } # => <script></script>
```

### After
```ruby
script { safe(%(console.log("Hi 👋"))) } # => <script>console.log("Hi 👋")</script>
```


joeldrapper:
Makes sense.

joeldrapper (approved):

PR #774, opened 2024-09-07
New testing module

joeldrapper:
Introducing a new testing module to replace the external gems `phlex-testing-capybara` and `phlex-testing-nokogiri`.

### DIFF
--- a/lib/phlex/testing/capybara.rb
+++ b/lib/phlex/testing/capybara.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require "capybara"
+
+module Phlex::Testing::Capybara
+	include Phlex::Testing::SGML
+
+	def self.included(mod)
+		if defined?(Minitest::Test) && Minitest::Test > mod
+			require "capybara/minitest"
+			include Capybara::Minitest::Assertions
+		end
+
+		if defined?(RSpec::Core::ExampleGroup) && RSpec::Core::ExampleGroup > mod
+			require "capybara/rspec"
+			include Capybara::RSpecMatchers
+		end
+	end
+
+	attr_reader :page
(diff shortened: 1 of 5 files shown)


lib/phlex/testing/capybara.rb
```diff
@@ -0,0 +1,28 @@
...
+	def self.included(mod)
+		if defined?(Minitest::Test) && Minitest::Test > mod
+			require "capybara/minitest"
+			include Capybara::Minitest::Assertions
+		end
+
+		if defined?(RSpec::Core::ExampleGroup) && RSpec::Core::ExampleGroup > mod
+			require "capybara/rspec"
+			include Capybara::RSpecMatchers
+		end
```
joeldrapper, line 17: Automatically detect which testing framework you’re using an include relevant matchers / assertions.

lib/phlex/testing/capybara.rb
```diff
@@ -0,0 +1,28 @@
...
+	end
+
+	attr_reader :page
+	alias_method :component, :page
```
joeldrapper, line 21: We need to have `page` for the built in capybara matchers, but you may want to refer to it as `component`.

lib/phlex/testing/nokogiri.rb
```diff
@@ -0,0 +1,19 @@
...
+		::Nokogiri::HTML5.fragment(
+			render_to_string(...),
+		)
+	end
+
+	def render_document(...)
+		::Nokogiri::HTML5(
+			render_to_string(...),
+		)
+	end
```
joeldrapper, line 18: In Nokogiri and Nokolexbor, you need to specify if you’re rendering a document or a fragment.

Issue #779, opened 2024-09-12
`mix` handles `nil`s poorly


davekaro:
Before Phlex 2, this worked:

```
test "string + nil" do
	output = mix({ class: "a" }, { class: nil })
	expect(output) == { class: "a" }
end
```

The functionality existed and was then restored in https://github.com/phlex-ruby/phlex/pull/705 and then broken again in https://github.com/phlex-ruby/phlex/pull/759/commits/0f8d8cbc6b919a4749674a6f68d7b426f5d97ecc#diff-1379affdd94a5363d3331840af91bc6171dac9338b2f6bd886a0636ac6258a38

PR #780, opened 2024-09-12
Fix `mix` to accept old or new nil values

davekaro:
Fixes https://github.com/phlex-ruby/phlex/issues/779, but ~I don't think it's all the way there because it only handles the `[String, nil]` combo~. I don't know the reasons the code changed for `mix` so much in https://github.com/phlex-ruby/phlex/commit/0f8d8cbc6b919a4749674a6f68d7b426f5d97ecc#diff-1379affdd94a5363d3331840af91bc6171dac9338b2f6bd886a0636ac6258a38, so this is more a starting point.

Update: in the second commit I made it work for the other types, but I'm sure there are other approaches to consider.

### DIFF
--- a/lib/phlex/helpers.rb
+++ b/lib/phlex/helpers.rb
@@ -26,8 +26,10 @@ module Phlex::Helpers
 					[old] + new.to_a
 				in [String, String]
 					"#{old} #{new}"
+				in [_, nil]
+					old
 				else
-					new.nil? ? old : new
+					new
 				end
 			end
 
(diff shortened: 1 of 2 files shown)


lib/phlex/helpers.rb
```diff
@@ -27,7 +27,7 @@ def mix(*args)
 				in [String, String]
 					"#{old} #{new}"
 				else
-					new
+					new.nil? ? old : new
```
willcosgrove, line 30: What do you think about following the pattern matching style with something like:

```rb
case [old, new]
# ...
in [_, nil]
  old
else
  new
end
```

davekaro: the `_` was exactly what I was looking for. Thanks!

Issue #784, opened 2024-09-19
Chopping block


joeldrapper:
Items in consideration for 2.0 chopping block:

### underscored method names
These could be replaced with `tag(:name)` though that feature will need re-implementing.

### yield_content
I think we can make `yield_content` private.

### view_context
We’ve already discussed renaming this `context`. We should really come up with a better name for Phlex internal context before we move forward with this.

### CSV helpers proxy
I think we should remove the helpers proxy for CSVs and instead allow `register_value_helper` in CSVs in Phlex-Rails.

### element clobbering guard
Without underscored methods, we wouldn’t need the clobbering guard.

### deferred render as a module
Instead of using a module for deferred render, I think it would be better to do something like

```ruby
def deferred = true
```

hangsu:
None of these changes would pose a big roadblock for Clearscope's codebase.

joeldrapper:
> None of these changes would pose a big roadblock for Clearscope's codebase.

Good to hear. Just trying to take advantage of the major release to get rid of a few things that have proven to be low-value.

hangsu:
Maybe this isn't the right place for it, but we're in the process of upgrading to 2.0 beta and our only complaint is the forced dasherizing of symbol values in `__attributes__`. Dasherizing of keys has always been fine.

I wish it was at least easily overridden.

hangsu:
Could there be a `format_attribute_value` method similar to `format_object` that can be overridden?

joeldrapper:
@hangsu I opened this discussion to talk about that. We already had a 3-day debate about it in Discord 😆  https://github.com/orgs/phlex-ruby/discussions/785

PR #786, opened 2024-09-23
Remove deferred render

joeldrapper:
This change removes the deferred render pattern as something you can opt into and instead makes it the default behaviour, with one small difference. You can `yield` any content that was captured from the component. It is no longer vanished.

### DIFF
--- a/lib/phlex/html/standard_elements.rb
+++ b/lib/phlex/html/standard_elements.rb
@@ -5,44 +5,83 @@ module Phlex::HTML::StandardElements
 	extend Phlex::SGML::Elements
 
 	# Outputs an `<a>` tag.
+	# The `<a>` element creates a hyperlink to web pages, files, email addresses, locations in the same page, or anything else a URL can address.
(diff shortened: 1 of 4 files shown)


hangsu:
Will this behavior be replaced with a `def deferred = true` as suggested in #784?

joeldrapper:
> Will this behavior be replaced with a def deferred = true as suggested in https://github.com/phlex-ruby/phlex/issues/784?

This is actually a sketch I wanted to run by @willcosgrove for how we could merge deferred render and normal rendering so you don’t need to pick.

stephannv:
@joeldrapper did this sketch failed? What are the plans about DeferredRender? I'm keeping an eye on this because my gem [Phlex::Slotable](https://github.com/stephannv/phlex-slotable) depends on DeferredRender.

joeldrapper:
Yeah, I think we’re going to have to keep things essentially the way they’ve always been. I was hoping to be able to merge them together but there are too many issues. Phlex becomes less flexible.

PR #792, opened 2024-09-25
safe values should be raw

Vagab:
@joeldrapper @willcosgrove 
so I made `html_safe` and `safe` consistent in this pr and also created `__implicit_output__` and made `plain` use old `__text__` instead so we can render `html_safe` as `raw` by default, unless it's `plain`. Does that look good?

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -120,7 +120,7 @@ class Phlex::SGML
 	end
 
 	def plain(content)
-		unless __text__(content)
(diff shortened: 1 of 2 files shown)


Vagab:
@joeldrapper @willcosgrove does this pr look like something you want? Or should I rollback the change to `__implicit_output__`  that allows to pass `safe` into a component without `raw` on the other side and just fix the order in `elements.rb`?

Vagab:
@joeldrapper ping

PR #794, opened 2024-09-29
Fix an issue with include multiple Phlex Kits

joeldrapper:


lib/phlex/kit.rb
```diff
@@ -3,36 +3,44 @@
 module Phlex::Kit
 	module LazyLoader
 		def method_missing(name, ...)
-			if name[0] == name[0].upcase && __phlex_kit_constants__.include?(name) && __get_phlex_kit_constant__(name) && methods.include?(name)
+			mod = self.class
```
joeldrapper, line 6: This now searches for the constant from the class rather than from the kit, since it may find the constant from another kit.

PR #795, opened 2024-10-01
Fiz Zeitwerk reloading

joeldrapper:


lib/phlex/kit.rb
```diff
@@ -52,6 +52,8 @@ def const_added(name)
 		if Class === constant && constant < Phlex::SGML
 			constant.include(self)
 
+			constant = nil
```
joeldrapper, line 55: We need the old constant to be GC’d here and not referenced by the closures below.

PR #800, opened 2024-10-03
Specify deferred mode via method

joeldrapper:
Rather than including `Phlex::DeferredRender`, you’ll need to specify deferred mode by defining `deferred?`.


willcosgrove:
I think we can close this in favor of #805

PR #805, opened 2024-10-03
Remove deferred render entirely

joeldrapper:
This abstraction is too confusing. If you understand the concept, you can still use it by yielding early.

```ruby
def view_template
  vanish { yield }

  # ...
end
```

This PR also changes the callbacks so `before_template` is given the block and can yield it.

```ruby
def before_template
  vanish { yield }
end
```

You can put this in a module called `DeferredRender` if you want. It’s up to you.

### DIFF
--- a/lib/phlex.rb
+++ b/lib/phlex.rb
@@ -9,7 +9,6 @@ module Phlex
 	autoload :CSV, "phlex/csv"
 	autoload :Callable, "phlex/callable"
 	autoload :Context, "phlex/context"
-	autoload :DeferredRender, "phlex/deferred_render"
 	autoload :DoubleRenderError, "phlex/errors/double_render_error"
 	autoload :Elements, "phlex/elements"
 	autoload :Error, "phlex/error"
--- a/lib/phlex/deferred_render.rb
+++ b/lib/phlex/deferred_render.rb
@@ -1,5 +1,0 @@
-# frozen_string_literal: true
-
-module Phlex::DeferredRender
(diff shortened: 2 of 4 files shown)


willcosgrove:
I like this change, I think this actually going to be easier to explain, even though it requires "more code" to use. Because you can **see** what's happening.

What do you think about also adding the block to the `after_template` just for consistency? I don't know what the actual usecase would be for someone yielding the block in an after template (there wasn't really a reason to want to in the `before_template` either, before this).

davekaro:
This works great for me. I had to adjust the `vanish` call to be `vanish { yield(self) if defined?(yield) }`. I think for sure you want to `yield(self)` otherwise you can't call methods on the component. The `if defined?(yield)` is because sometimes I don't pass a block to customize the render.

willcosgrove:
@davekaro I had wondered about that as well. I think this works nicely and is succinct:

```rb
def before_template(&)
  vanish(&)
  super
end
```

`super` isn't strictly necessary, but it will let any other modules or parent classes continue with any hooks they might have installed.

davekaro:
Even better!

svyatogor:
This is not mentioned in the upgrade documentation. As I was relying on this feature took me a while to find this issue.

willcosgrove:
Thanks @svyatogor for pointing that out, sorry you were the one to have to run into that.

I've opened a PR to add it to the docs phlex-ruby/phlex.fun#55

PR #808, opened 2024-10-04
make vanish public API so people can add deferred render behavior if desired

davekaro:
Opening this as a discussion point, do you want to make `vanish` public API since `Phlex::DeferredRender` is gone now?

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -229,8 +229,6 @@ class Phlex::SGML
 		nil
 	end
 
-	private
-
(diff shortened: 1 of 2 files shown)


joeldrapper:
We renamed it from `__vanish__` to `vanish` to indicate that it is part of Phlex’ public API. Do you think it should be public on the component too?

davekaro:
Ahh yes that’s right. No I think it’s good as is, or at least I have no need for it to be public on the component.

Do we still want the tests for vanish? I could make some more if yes

PR #809, opened 2024-10-06
Add basic testing

SethHorsley:
now you can use 
```ruby
include Phlex::Testing::Basic
```
or
```ruby
include Phlex::Testing::ViewHelper
```


SethHorsley:
@joeldrapper was this merged or is there a way do testing now?

PR #811, opened 2024-10-10
Expose `styles` method

stephannv:
What do you think about exposing `styles` method? I was building some style for my form builder (generating HTML via Rails), and I could not use `style: { color: "white" }` because it isn't processed by Phlex.
```ruby
def view_template
  form_with(...) do |form|
    form.submit style: { background_color: "white", color: "red" }
    # => <input ... style="background_color white color red">
  end
end
```

So I thought using the same method Phlex uses but I noticed the `__styles__` method is "internal" and generates unsafe code. I think it would be nice to expose it to users. I'm not sure about the method name, and I think we need at least one test to ensure this method is exposed.

With this change, we can do something like:
```ruby
form.submit style: styles({ background_color: "white", color: "red" })
# => <input ... style="background_color: white; color: red">
```

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -267,6 +267,10 @@ class Phlex::SGML
 		nil
 	end
 
+	def styles(value)
+		__styles__(value).gsub('"', "&quot;")
+	end
+
 	def __yield_content__
 		return unless block_given?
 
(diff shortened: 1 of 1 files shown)


joeldrapper:
The method generates unsafe code because it’s escaped later. In theory, Rails should be escaping too. I’m not against making this part of our public API as long as it doesn’t introduce any security issues.

I recommend you add an `alias_method :styles, :__styles__` and get a feel for it for a few days and report back.

stephannv:
It makes sense @joeldrapper. I will close this PR for now.

PR #817, opened 2024-10-29
Detect invalid `href` attributes

### LINKED ISSUE #816: If you pass a Hash to `href`, you get an error about calling `downcase` on it

joeldrapper:
Closes #816

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -401,9 +401,16 @@ class Phlex::SGML
 			unless Phlex::SGML::SafeObject === v
 				normalized_name = lower_name.delete("^a-z-")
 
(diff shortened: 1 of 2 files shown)


joelmoss:
Not tried this yet, but [phlexible](https://github.com/joelmoss/phlexible/blob/master/lib/phlexible/rails/a_element.rb) overrides the `a` method in order to allow passing Rails route helpers to the href. Basically in the same way that the Rails `link_to` helper passes it's URL argument to the `url_for` helper.

So I can do things like...

```ruby
a(href: :some_route) { 'click me' }
a(href: article) { 'click me' }
a(href: [user, :articles]) { 'click me' }
```

I'm hoping that this change will not break the above 🤞

joeldrapper:
It should be fine if you’re only passing a string up to `super`.

PR #820, opened 2024-11-07
Renames BlackHole to Vanish for consistency

bradgessler:
This commit tidies up the API by making the naming consistent with https://github.com/phlex-ruby/phlex/pull/805


joeldrapper:
I think this makes sense.

PR #824, opened 2024-11-26
Update CI

joeldrapper:


phlex.gemspec
```diff
@@ -12,7 +12,7 @@ Gem::Specification.new do |spec|
 	spec.description = "A high-performance view framework optimised for fun."
 	spec.homepage = "https://www.phlex.fun"
 	spec.license = "MIT"
-	spec.required_ruby_version = ">= 3.3.1"
+	spec.required_ruby_version = ">= 3.1"
```
stephannv, line 15: How `const_added` would work on Ruby < 3.2?
Ref: https://rubyreferences.github.io/rubychanges/3.2.html#moduleconst_added

joeldrapper: You’re right, we need 3.2.

PR #825, opened 2024-11-26
Add fiber isolation

joeldrapper:
Thread locals are not Fiber-isolated and neither are Fiber locals Thread-isolated. The approach taken in this PR is to use a Thread local, storing the current fiber ID with the value. Then we can check the ids match when reading.


fcheung:
I won’t pretend to understand the code here, but given that you’re requiring Ruby 3.2 could you use Fiber[] (fiber storage - per fiber but inherited from parent fibers/threads) ?

joeldrapper:
As far as I can tell, `Fiber[]` does not give you Thread isolation.

PR #831, opened 2024-12-16
Assert style tests in Quickdraw

joeldrapper:
Convert all the Quickdraw tests to use assert style.


joeldrapper:
Looks like the TruffleRuby issue is unrelated. It’s a problem with Sus.

PR #834, opened 2025-01-30
Fragment caching

joeldrapper:
This PR introduces a new helper method to support fragment caching. I’ve been reluctant to add fragment caching because it’s difficult to expire the cache when templates change and I didn’t want to try to build up a static dependency tree between components, partials, etc.

For now, we’re keeping it simple by expiring the cache on each deploy via a deploy key (see below).

**Example:**
```ruby
cache @products do
  @products.each do |product|
    cache product do
      h1 { product.name }
    end
  end
end
```

The `cache` method will take user cache keys and combine them with built-in supplemental keys to cache the captured content against. `cache` will call `cache_store`, which must return a Phlex cache store.

## Supplemental keys

The `cache` method supplements your cache keys with the following:

1. `Phlex::DEPLOY_KEY` — the time that the app was booted and Phlex was loaded.
5. The name of the class where the caching is taking place. This prevents collisions between classes.
6. The name of the method where the caching is taking place. This prevents collisions between different methods in the same class.
7. The line number where the `cache` method is called. This prevents collisions between different lines, especially when no custom cache keys are provided.

## Low level caching

The `low_level_cache` method requires a cache key from you and you control the entire cache key.

## Cache store interface

Cache stores are objects that respond to `fetch(key, &content)`. This method must return the result of `yield` or a previously cached result of `yield`. This method may raise if the result of `yield` is not a string or if the key is not valid.

It’s up to you what keys are valid, but note that Phlex itself uses arrays, string and integers in its keys.

This interface is a subset of Rails’ cache interface, meaning Rails cache interface implements this interface can can be used as a Phlex cache store.

## `FIFOCacheStore`

This PR also introduces a new cache store based on the `Phlex::FIFO`. This is an extremely fast in-memory cache store that evicts keys on a first-in-first-out basis. It can be initialised with a max bytesize.

### DIFF
--- a/lib/phlex.rb
+++ b/lib/phlex.rb
@@ -28,7 +28,6 @@ module Phlex
 
 	CACHED_FILES = Set.new
 	DEPLOY_KEY = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
-	CACHE_STORE = NullCacheStore.new
 
 	def self.__expand_attribute_cache__(file_path)
 		unless CACHED_FILES.include?(file_path)
--- a/lib/phlex.rb
+++ b/lib/phlex.rb
@@ -17,17 +17,16 @@ module Phlex
 	autoload :Helpers, "phlex/helpers"
 	autoload :Kit, "phlex/kit"
 	autoload :NameError, "phlex/errors/name_error"
-	autoload :NullCacheStore, "phlex/null_cache_store"
 	autoload :SGML, "phlex/sgml"
 	autoload :SVG, "phlex/svg"
 	autoload :Vanish, "phlex/vanish"
 
 	Escape = ERB::Escape
-	ATTRIBUTE_CACHE = FIFO.new
 	Null = Object.new.freeze
 
-	CACHED_FILES = Set.new
 	DEPLOY_KEY = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
+	CACHED_FILES = Set.new
+	ATTRIBUTE_CACHE = FIFO.new
 
 	def self.__expand_attribute_cache__(file_path)
 		unless CACHED_FILES.include?(file_path)
--- a/lib/phlex.rb
+++ b/lib/phlex.rb
@@ -5,7 +5,6 @@ require "set"
 
 module Phlex
 	autoload :ArgumentError, "phlex/errors/argument_error"
-	autoload :Vanish, "phlex/vanish"
 	autoload :CSV, "phlex/csv"
 	autoload :Callable, "phlex/callable"
 	autoload :Context, "phlex/context"
@@ -13,18 +12,23 @@ module Phlex
 	autoload :Elements, "phlex/elements"
 	autoload :Error, "phlex/error"
 	autoload :FIFO, "phlex/fifo"
+	autoload :FIFOCacheStore, "phlex/fifo_cache_store"
 	autoload :HTML, "phlex/html"
 	autoload :Helpers, "phlex/helpers"
 	autoload :Kit, "phlex/kit"
 	autoload :NameError, "phlex/errors/name_error"
+	autoload :NullCacheStore, "phlex/null_cache_store"
 	autoload :SGML, "phlex/sgml"
 	autoload :SVG, "phlex/svg"
+	autoload :Vanish, "phlex/vanish"
 
 	Escape = ERB::Escape
 	ATTRIBUTE_CACHE = FIFO.new
 	Null = Object.new.freeze
 
 	CACHED_FILES = Set.new
+	DEPLOY_KEY = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
+	CACHE_STORE = NullCacheStore.new
 
 	def self.__expand_attribute_cache__(file_path)
 		unless CACHED_FILES.include?(file_path)
--- a/lib/phlex/fifo.rb
+++ b/lib/phlex/fifo.rb
@@ -7,7 +7,7 @@ class Phlex::FIFO
 		@max_bytesize = max_bytesize
 		@max_value_bytesize = max_value_bytesize
 		@bytesize = 0
-		@mutex = Mutex.new
+		@mutex = Monitor.new
 	end
 
 	attr_reader :bytesize, :max_bytesize
--- a/lib/phlex/fifo_cache_store.rb
+++ b/lib/phlex/fifo_cache_store.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+class Phlex::FIFOCacheStore
+	def initialize(max_bytesize: 2 ** 20)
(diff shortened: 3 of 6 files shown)


lib/phlex/fifo.rb
```diff
@@ -7,7 +7,7 @@ def initialize(max_bytesize: 2_000, max_value_bytesize: 2_000)
 		@max_bytesize = max_bytesize
 		@max_value_bytesize = max_value_bytesize
 		@bytesize = 0
-		@mutex = Mutex.new
+		@mutex = Monitor.new
```
joeldrapper, line 10: We need this to be a Monitor to avoid deadlocks.

lib/phlex.rb
```diff
@@ -5,26 +5,30 @@
...
 	autoload :SVG, "phlex/svg"
+	autoload :Vanish, "phlex/vanish"
 
 	Escape = ERB::Escape
 	ATTRIBUTE_CACHE = FIFO.new
 	Null = Object.new.freeze
 
 	CACHED_FILES = Set.new
+	DEPLOY_KEY = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
+	CACHE_STORE = NullCacheStore.new
```
joeldrapper, line 31: This needs to be a configuration option, not a constant.

lib/phlex/sgml.rb
```diff
@@ -223,6 +223,19 @@ def render(renderable = nil, &)
 		nil
 	end
 
+	def cache(*keys, **, &)
+		context = @_context
+		return if context.fragments && !context.in_target_fragment
+
+		context.buffer << cache_store.fetch(
+			[Phlex::DEPLOY_KEY, self.class.name, __method__, keys].freeze, **
```
joeldrapper, line 231: Breaking this down:

1. `Phlex::DEPLOY_KEY` — this is set to the current time on boot and is necessary for invalidating the cache on deploy in case you’ve changed a Phlex component.
2. `self.class.name` — this isolates the cache to the current class meaning you don’t need to worry about cross-class conflicts.
3. `__method__` — this isolates the cache to the current method, e.g. `view_template` so you don’t need to worry about using different keys in different methods on the same class.
4. `keys` finally, we use your own keys.

lib/phlex/sgml.rb
```diff
@@ -223,6 +223,19 @@ def render(renderable = nil, &)
...
+		) { capture(&) }
+	end
+
+	def cache_store
+		Phlex::CACHE_STORE
+	end
```
joeldrapper, line 237: We go via the `cache_store` method so you can use different cache stores for different components.

lib/phlex/sgml.rb
```diff
@@ -223,6 +223,19 @@ def render(renderable = nil, &)
 		nil
 	end
 
+	def cache(*keys, **, &)
```
joeldrapper, line 226: We should require at least one key here. Alternatively, if no key is provided, we could use `caller_locations[0].lineno`.

lib/phlex/fifo_cache_store.rb
```diff
@@ -0,0 +1,40 @@
...
+		if (result = fifo[key])
+			result
+		else
+			result = yield
+			fifo[key] = result
+			result
+		end
+	end
+
+	def map_key(value)
```
joeldrapper, line 24: I wonder if this should keep track of the rough bitesize of the full key and raise if it gets too large, since the objects will never be GC’d while they’re in the cache. 🤔

willcosgrove (approved):
This looks good to me! Excited to try it out in my `Icon` component

lib/phlex/fifo_cache_store.rb
```diff
@@ -0,0 +1,40 @@
...
+		case value
+		when Array
+			value.map { |it| map_key(it) }
+		when Hash
+			value.to_h { |k, v| [map_key(k), map_key(v)].freeze }
```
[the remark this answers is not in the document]
joeldrapper: Yes, it means Ruby can allocate an exact sized array.

lib/phlex.rb
joeldrapper: Actually, we can use a method on the component for this configuration.

lib/phlex/sgml.rb
```diff
@@ -223,6 +223,58 @@ def render(renderable = nil, &)
...
+		return if context.fragments && !context.in_target_fragment
+
+		context.buffer << cache_store.fetch(cache_key, **options) { capture(&content) }
+	end
+
+	# Points to the cache store used by this component.
+	# By default, it points to `Phlex::NullCacheStore`, which does no caching.
+	# Override this method to use a different cache store.
+	def cache_store
+		Phlex::NullCacheStore
```
joeldrapper, line 275: I actually wonder if this shouldn’t raise an error so you know you need to configure a cache store.

jmdfm:
I see the note about `low_level_cache` but how can I manually expire a cache key? It would be great if this was exposed to the developer beyond doing a redeploy, because that's much more powerful to me. 

Here's my use case: I have a view with 7 columns for each day of the week, and each column lists tasks. There is a repeated task that occurs on 2 of the 7 days, and you change its title. I'd want to be able to manually expire those two cache keys for the affected days, and redraw the view.

Right now I'm having to redraw all 7 days, which is not great and would be otherwise trivial to russian doll cache if I was using ERB.

joeldrapper:
@jmdfm you can pass your own keys to the `cache` method. I’m not sure I follow your example, but if any of the keys you pass to `cache` change, it will yield the block and cache it against those new keys.

jmdfm:
Ok right, so if the keys change, the block is yielded and a re-cache happens. This doesn't seem to take into account if any of the attributes of the cached object change and should yield and re-cache?

Simplifying my example: you have 7 columns of todos, and the user updates the text of one of the todos. If we're only looking at cache keys to know when to yield and re-cache, to me it seems like this implementation will not show that change made by the user?

Rails' cache mechanism has a `template tree digest` in the cache key so that if the object changes, the cache is automatically invalidated: https://guides.rubyonrails.org/caching_with_rails.html#fragment-caching: 

```
The string of characters in the middle is a template tree digest. It is a hash digest computed based on the contents of the view fragment you are caching. If you change the view fragment (e.g., the HTML changes), the digest will change, expiring the existing file.
```

joeldrapper:
Rails template tree digest solves a different problem, the one that we solve by just expiring all the keys on deploy. The way that rails handles nested caches (Russian doll caching) is via child models touching their parent models when they are updated. See this Rails guide. https://guides.rubyonrails.org/caching_with_rails.html

Our implementation should be compatible with that since we will call .cache_key on the model, which will use its modification date as part of the key.

jmdfm:
Ah great, missed that in the diff view. 💪

joeldrapper:
We will either need to find a solution to selective rendering or accept #836 before continuing with this.

PR #842, opened 2025-02-03
Add `SGML#fragment` for marking selective rendering blocks

willcosgrove:
- [x] Implement `SGML#fragment` method
- [x] Support fragments inside of a cached block

### DIFF
--- a/lib/phlex/context.rb
+++ b/lib/phlex/context.rb
@@ -8,6 +8,7 @@ class Phlex::Context
 		@user_context = user_context
 		@fragments = nil
 		@fragment_depth = 0
+		@cache_stack = []
 		@halt_signal = nil
 		@view_context = view_context
 	end
@@ -17,7 +18,7 @@ class Phlex::Context
 	attr_reader :fragments, :fragment_depth, :view_context
 
 	def target_fragments(fragments)
-		@fragments = fragments.to_h { |it| [it, true].freeze }
+		@fragments = fragments.to_set
 	end
 
 	def around_render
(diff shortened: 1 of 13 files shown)


lib/phlex/context.rb
```diff
@@ -6,24 +6,59 @@ def initialize(user_context: {}, view_context: nil)
...
 		@view_context = view_context
 	end
 
 	attr_accessor :buffer, :capturing, :user_context
 
-	attr_reader :view_context
+	attr_reader :fragments, :fragment_depth, :view_context
+
+	def target_fragments(fragments)
+		@fragments = fragments.to_h { |it| [it, true].freeze }
```
joeldrapper, line 20: Let’s just use a set here unless there’s a big performance impact.

lib/phlex/sgml.rb
```diff
@@ -167,6 +176,13 @@ def capture(*args, &block)
 		end
 	end
 
+	# Define a named fragment that can be selectively rendered.
+	def fragment(name)
+		@_context.begin_target(name)
+		yield
+		@_context.end_target(name)
+	end
```
joeldrapper, line 184: This method should return `nil`.

lib/phlex/context.rb
```diff
@@ -6,24 +6,59 @@ def initialize(user_context: {}, view_context: nil)
...
+	def begin_target(id)
+		@fragment_depth += 1 if @fragments&.[](id)
+	end
+
+	def end_target(id)
+		return unless @fragments&.[](id)
+
+		@fragments.delete(id)
+		@fragment_depth -= 1
+		throw @halt_signal if @fragments.length == 0
```
joeldrapper, line 45: This is how we had it before, but throwing `nil` is odd here. I think we can rework this a bit to make it better.

PR #855, opened 2025-02-13
Allow `Phlex::HTML` and `Phlex::SVG` to be rendered inline without a class

marcoroth:
It's quite common in Rails' View Helpers to implement `content_tag` or `tag.div` calls and chain them using `+`, `concat`, or `safe_content`.

Since the Phlex DSL is kinda expressive in that sense we can expose `Phlex.html` and `Phlex.svg` to inline render HTML/SVGs to a String without having the need to create a class for it, making this ideal for using Phlex in regular Rails View Helpers.

Here are some examples on how you might be able to refactor this using `Phlex.html`:

<details>
<summary>
Redmine's <code>ApplicationHelper</code>
</summary>

[[Source]](https://github.com/redmine/redmine/blob/bc90e0670df930b0a4aa8be1e0d887aac45eea0f/app/helpers/application_helper.rb#L856-L867)

**Before:**
```ruby
module ApplicationHelper
  def actions_dropdown(&)
    content = capture(&)
    if content.present?
      trigger =
        content_tag('span', sprite_icon('3-bullets', l(:button_actions)), :class => 'icon-only icon-actions',
                    :title => l(:button_actions))
      trigger = content_tag('span', trigger, :class => 'drdn-trigger')
      content = content_tag('div', content, :class => 'drdn-items')
      content = content_tag('div', content, :class => 'drdn-content')
      content_tag('span', trigger + content, :class => 'drdn')
    end
  end
end
```

**After:**
```ruby
module ApplicationHelper
  def actions_dropdown(&)
    content = capture(&)

    if content.present?
      Phlex.html do
        span(class: 'drdn') do
          span(class: 'drdn-trigger') do
            span(class: 'icon-only icon-actions', title: l(:button_actions)) do
              sprite_icon('3-bullets', l(:button_actions))
            end
          end

          div(class: 'drdn-content') do
            div(class: 'drdn-items') do
              raw(content)
            end
          end
        end
      end
    end
  end
end
```

</details>

<details>
<summary>
GitLab's <code>Tabs</code>
</summary>

[[Source]](https://github.com/gitlabhq/gitlabhq/blob/9bb15d0553a2a66bf9feb5e513ec79c2424c8f1c/spec/frontend/fixtures/tabs.rb#L18-L22)

**Before:**
```ruby
panels = content_tag(:div, class: 'tab-content') do
  content_tag(:div, 'Foo', { id: 'foo', class: 'tab-pane active', data: { testid: 'foo-panel' } }) +
[...]

PR #860, opened 2025-02-14
Build HTML and SVG with an inline DSL

joeldrapper:
@marcoroth here’s a slightly different take on #855, which I think is slightly simpler to understand and explain.

1. Instance variables are copied over to the component (unless they are `@_state` or `@_content_block` which Phlex uses internally)
2. The block is executed in the component context via `instance_exec`, but is still able to access local variables.
3. If you need to call a method that is not a Phlex method, you can take a yield param and call the method on that.


lib/phlex.rb
```diff
@@ -43,6 +43,38 @@ def self.eager_load
...
+	def self.html(&block)
+		HTML.call do |component|
+			receiver = block.binding.receiver
+
+			receiver.instance_variables.each do |ivar|
+				value = receiver.instance_variable_get(ivar)
+
+				unless component.instance_variable_defined?(ivar)
+					component.instance_variable_set(ivar, value)
+				end
```
willcosgrove, line 56: We can skip early to avoid accessing the ivar if we're not going to assign it. (Don't merge this suggestion, I've got spaces instead of tabs here 🙈)

```suggestion
        next if component.instance_variable_defined?(ivar)
        
				value = receiver.instance_variable_get(ivar)
				component.instance_variable_set(ivar, value)
```

willcosgrove:
This is great! I hadn't even considered it could work without a "proxy" component. Clever!

PR #862, opened 2025-02-15
Restore access to context in `render?` hook

willcosgrove:


joeldrapper:
Thanks for catching this! It’s now in the latest release `2.0.1`.

PR #863, opened 2025-02-16
Raise a helpful error if you access `context` before rendering

joeldrapper:
If you access the `context` before rendering, the error message says `nil` doesn’t respond to `user_context`. This PR makes `context` return `nil` instead. Opening as a draft as I want to write a test for this before merging.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -94,7 +94,19 @@ class Phlex::SGML
 	end
 
 	def context
-		@_state.user_context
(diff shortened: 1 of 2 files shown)


joeldrapper:
It might even make sense for this to raise a helpful error explaining why you can't access the context before rendering.

But we need a way to check if we're rendering or not. Maybe checking if the context is nil isn't the best way to do that.

I wonder if we should also have some sort of `rendering?` method. It could just be `!!@_state`.

PR #866, opened 2025-02-18
Faster CSVs

joeldrapper:


lib/phlex/csv.rb
```diff
@@ -1,64 +1,95 @@
...
-			raise <<~MESSAGE
-				You need to define escape_csv_injection? in #{self.class.name}, returning either `true` or `false`.
+		ensure_escape_csv_injection_configured!
 
-				CSV injection is a security vulnerability where malicious spreadsheet formulae are used to execute code or exfiltrate data when a CSV is opened in a spreadsheet program such as Microsoft Excel or Google Sheets.
+		strip_whitespace = trim_whitespace?
+		escape_csv_injection = escape_csv_injection?
+		delimiter = @delimiter
+		row_buffer = @_row_buffer
+		has_yielder = respond_to?(:yielder, true)
```
joeldrapper, line 27: Let’s do a warning here if you have a yielder. We don’t need a yielder and wrapper.

lib/phlex/csv.rb
```diff
@@ -100,22 +120,99 @@ def trim_whitespace?
...
+
+				if FORMULA_PREFIXES_MAP[first_byte]
+					buffer << '"\'' << value.gsub('"', '""') << '"'
+				elsif value.match?(/^\s|\s$|[\n",]/)
+					buffer << '"' << value.gsub('"', '""') << '"'
+				else
+					buffer << value
+				end
+			else # not escaping CSV injection
+				if value.match?(/^\s|\s$|[\n",]/)
```
joeldrapper, line 164: These regexes should include the `delimiter`. We’ll have to generate them at the top and pass them down.

lib/phlex/csv.rb
```diff
@@ -1,64 +1,95 @@
...
 	def call(buffer = +"", context: nil)
-		unless escape_csv_injection? == true || escape_csv_injection? == false
-			raise <<~MESSAGE
-				You need to define escape_csv_injection? in #{self.class.name}, returning either `true` or `false`.
+		ensure_escape_csv_injection_configured!
 
-				CSV injection is a security vulnerability where malicious spreadsheet formulae are used to execute code or exfiltrate data when a CSV is opened in a spreadsheet program such as Microsoft Excel or Google Sheets.
+		strip_whitespace = trim_whitespace?
+		escape_csv_injection = escape_csv_injection?
+		delimiter = @delimiter
```
joeldrapper, line 25: We should probably only allow certain characters as delimiters.

lib/phlex/csv.rb
```diff
@@ -100,22 +120,99 @@ def trim_whitespace?
...
+			view_template(...)
+		else
+			super
+		end
+	end
+
+	def ensure_escape_csv_injection_configured!
+		if escape_csv_injection? == UNDEFINED
+			raise <<~MESSAGE
+				You need to define escape_csv_injection? in #{self.class.name}, returning either `true` or `false`.
```
joeldrapper, line 196: Technically this only needs to be truthy or falsy.

lib/phlex/csv.rb
```diff
@@ -100,22 +120,99 @@ def trim_whitespace?
...
+
+				CSV injection is a security vulnerability where malicious spreadsheet formulae are used to execute code or exfiltrate data when a CSV is opened in a spreadsheet program such as Microsoft Excel or Google Sheets.
```
joeldrapper, line 198: Let’s wrap this at a reasonable width.

lib/phlex/csv.rb
```diff
@@ -1,64 +1,95 @@
...
-			yielder(record) do |*args, **kwargs|
-				view_template(*args, **kwargs)
+			i = 0
+			number_of_columns = row.length
+			first_col = true
 
-				if @_first && render_headers?
-					buffer << @_headers.join(@delimiter) << "\n"
+			while i < number_of_columns
+				header, value = row[i]
```
joeldrapper, line 70: We need to check that this header isn’t inconsistent.

PR #867, opened 2025-02-19
[EXPERIMENTAL] Add support for ERB snippets and sidecar templates

joeldrapper:
This PR adds experimental support for ERB snippets in Phlex components.

```ruby
class Nav < Phlex::HTML
  erb :view_template, <<~ERB
    <nav>
      <% yield %>
    </nav>
  ERB
  
  erb :item, <<~ERB, locals: %(href:)
    <a href="<%= href %>">
      <% yield %>
    </a>
  ERB
end
```

Now this component can be called like this from Phlex.

```ruby
class Example < Phlex::HTML
  def view_template
    Nav do |n|
      n.item(href: "/") { "Home" }
    end
  end
end
```

Or from ERB in Phlex

```ruby
class Example < Phlex::HTML
  erb :view_template, <<~ERB
    <% Nav do |n| %>
      <% n.item(href: "/") do %>
        Home
      <% end %>
    <% end %>
  ERB
end
```

## What is a snippet?

Any instance method in a Phlex component is a snippet. You can use the `erb` macro instead of defining a method. If you define a snippet called `view_template`, that will be the main template for the component.

## Inline templates

You can define the ERB templates inline like this

```ruby
erb :snippet_name, <<~ERB
  <h1>Hello</h1>
ERB
```

## External templates

You can define ERB templates that will be loaded from external template files.

```ruby
erb :snippet_name
```

If your component was called `card.rb`, this would read `card/snippet_name.html.erb` as your ERB template. There’s a special case for the `view_template` snippet. In this case, if the file `card/view_template.html.erb` can’t be found, it will look for `card.html.erb` next to the original `card.rb`.

### Locals

You can define parameters in a string passed as the third positional argument.

```ruby
erb :view_template, locals: "(req, opt = nil, keyreq:, keyopt: nil)"
```

This string works as if you had used it in your method definition

```ruby
def view_template(req, opt = nil, keyreq:, keyopt: nil)
  #              ^                                    ^
  #              ^------------------------------------^
  #                              here
```

We recommend using `%()` strings because they look more like params

```ruby
erb :view_template, locals: %(req, opt = nil, keyreq:, keyopt: nil)
```

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -5,6 +5,17 @@ class Phlex::SGML
 	UNSAFE_ATTRIBUTES = Set.new(%w[srcdoc sandbox http-equiv]).freeze
 	REF_ATTRIBUTES = Set.new(%w[href src action formaction lowsrc dynsrc background ping]).freeze
 
+	ERBCompiler = ERB::Compiler.new("<>").tap do |compiler|
+		compiler.pre_cmd    = [""]
+		compiler.put_cmd    = "@_state.buffer.<<"
+		compiler.insert_cmd = "__implicit_output__"
+		compiler.post_cmd   = ["nil"]
+
+		def compiler.add_insert_cmd(out, content)
+			out.push("#{@insert_cmd}((#{content}))")
+		end
+	end
+
 	autoload :Elements, "phlex/sgml/elements"
 	autoload :SafeObject, "phlex/sgml/safe_object"
 	autoload :SafeValue, "phlex/sgml/safe_value"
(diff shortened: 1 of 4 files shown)


lib/phlex/sgml.rb
```diff
@@ -5,6 +5,17 @@ class Phlex::SGML
...
 	REF_ATTRIBUTES = Set.new(%w[href src action formaction lowsrc dynsrc background ping]).freeze
 
+	ERBCompiler = ERB::Compiler.new("<>").tap do |compiler|
+		compiler.pre_cmd    = [""]
+		compiler.put_cmd    = "@_state.buffer.<<"
+		compiler.insert_cmd = "__implicit_output__"
+		compiler.post_cmd   = ["nil"]
+
+		def compiler.add_insert_cmd(out, content)
+			out.push("#{@insert_cmd}((#{content}))")
```
joeldrapper, line 15: We define this because the original implementation called `to_s` on everything which meant we couldn’t have different behaviour for different things.

joeldrapper:
Closing this for now.

PR #876, opened 2025-03-07
Adds and passes test for blank and nil csv values

nolantait:
Currently, calling a CSV template with blank or `nil` values fails when looking up the first byte:

```ruby
# lib/phlex/csv.rb

if FORMULA_PREFIXES_MAP[value.getbyte(0)]

# ...

first_byte = value.getbyte(0)
if FORMULA_PREFIXES_MAP[first_byte]
```

Raises a confusing `eval error: no implicit conversion from nil to integer`

This PR adds and passes specs that match the options for stripping whitespace.

Its currently written to do either:

- `""` for blank values if not stripping whitespace, Ruby parses them to `""`
- Nothing when stripping whitespace, Ruby parses them it to `nil`

### DIFF
--- a/lib/phlex/csv.rb
+++ b/lib/phlex/csv.rb
@@ -168,7 +168,9 @@ class Phlex::CSV
 			value = value.strip
 
 			if escape_csv_injection
-				if FORMULA_PREFIXES_MAP[value.getbyte(0)]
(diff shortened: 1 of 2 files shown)


joeldrapper:
Closes #878

PR #880, opened 2025-03-17
Move capturing up to the state object

joeldrapper:
In order to fix https://github.com/phlex-ruby/phlex-rails/issues/276, we need to simplify the capturing interface so it can be swapped out. This PR removes the ability to capture into an arbitrary buffer. Instead, it depends on reading the new buffer from the state within a capture block.

See https://github.com/phlex-ruby/phlex-rails/pull/279


lib/phlex/sgml.rb
```diff
@@ -353,15 +353,8 @@ def enable_cache_reloading?
...
-		if args.length > 0
-			@_state.capturing_into(Phlex::Vanish) { yield(*args) }
-		else
-			@_state.capturing_into(Phlex::Vanish) { yield(self) }
-		end
-
+	def vanish(...)
+		capture(...)
 		nil
 	end
```
joeldrapper, line 359: I think this is a reasonable simplification here. It allocates a string but I think it’s fine.

lib/phlex/sgml/state.rb
```diff
@@ -14,16 +14,7 @@ def initialize(user_context: {}, output_buffer:, fragments:)
...
-	attr_reader :fragments, :fragment_depth, :output_buffer
-
-	def buffer
-		case @buffer
-		when Proc
-			@buffer.call
-		else
-			@buffer
-		end
-	end
```
joeldrapper, line 26: We never needed this.

lib/phlex/sgml.rb
willcosgrove: Yeah I think so. The typical use case for `vanish` is that you'd be calling it with a block that _shouldn't_ contain any direct element calls. So the performance optimization of `Phlex::Vanish` probably wasn't even being used much.

willcosgrove (approved):
This looks good to me! `capture` still maintains its same contract so this can be a non-breaking update 💪

peterberkenbosch:
<3

PR #881, opened 2025-03-18
Update around_row to forward all arguments to row_template

willcosgrove:
Previously, one of the uses of the `yielder` override in `Phlex::CSV` was to change the argument signature provided to the `row_template` method. The new `around_row` was not allowing for any modifications to the argument structure.

### DIFF
--- a/lib/phlex/csv.rb
+++ b/lib/phlex/csv.rb
@@ -113,8 +113,8 @@ class Phlex::CSV
 		buffer
 	end
 
-	def around_row(item)
-		row_template(item)
+	def around_row(...)
+		row_template(...)
 	end
 
 	def filename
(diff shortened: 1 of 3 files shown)


lib/phlex/csv.rb
```diff
@@ -113,8 +113,8 @@ def call(buffer = +"", context: nil, delimiter: self.delimiter)
 		buffer
 	end
 
-	def around_row(item)
-		row_template(item)
+	def around_row(...)
+		row_template(...)
```
joeldrapper, line 117: This looks good, but the tests are failing.

joeldrapper (approved):

lib/phlex/csv.rb
willcosgrove: Oh weird, I'll take a look. They're passing locally

PR #889, opened 2025-04-04
Root hash attributes

### LINKED ISSUE #887: Support root hash property with flattened hash attributes
```ruby
div(data: { _: "root", foo: "bar" })
```

```html
<div data="root" data-foo="bar'></div>
```

We should also update `mix` to support this.

```ruby
in [_, Hash]
  { _: old, **new }
in [Hash, _]
  { **old, _: new }
```

willcosgrove:
This closes #887 by allowing hash attribute values to set the root attribute using the `:_` key.

```rb
div(data: { _: "root", controller: "copy" })
```

```html
<div data="root" data-controller="copy"></div>
```

It also includes a change to `mix` to allow it to mix scalar values with hash values.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -566,7 +566,7 @@ class Phlex::SGML
 		attributes.each do |k, v|
 			next unless v
 
-			if k == :_
+			if (root_key = (:_ == k))
 				name = ""
 				original_base_name = base_name
 				base_name = base_name.delete_suffix("-")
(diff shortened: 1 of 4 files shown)


lib/phlex/sgml.rb
```diff
@@ -566,14 +566,20 @@ def __nested_attributes__(attributes, base_name, buffer = +"")
 		attributes.each do |k, v|
 			next unless v
 
-			name = case k
-				when String then k
-				when Symbol then k.name.tr("_", "-")
-				else raise Phlex::ArgumentError.new("Attribute keys should be Strings or Symbols")
-			end
+			if k == :_
```
willcosgrove, line 569: Given that we have to check this twice, it may be faster to store the result in a local
```suggestion
			if root_key = (k == :_)
```

willcosgrove: It may also be better to flip this equality check around since we always know what `:_` is.
```suggestion
			if :_ == k
```

PR #897, opened 2025-04-21
Require version when loading gem

fusion2004:
Apps and other gems (like lookbook) want to be able to easily tell whether they are working with Phlex v1 or v2, and lookbook is now trying to check `Phlex::VERSION` for that, which is not loaded by default.

Would it make sense to always load that? It seems like other major gems I checked, like sidekiq, do.

### DIFF
--- a/lib/phlex.rb
+++ b/lib/phlex.rb
@@ -3,6 +3,7 @@
 require "erb"
 require "set"
 require "zeitwerk"
+require "phlex/version"
 
 module Phlex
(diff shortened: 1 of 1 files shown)


stephannv:
Which version are you using? It seems that after 2.1 the version is always loaded, If you are < 2.1, you need to `require "phlex/version"`.
Screenshot 2025-04-21 at 11 39 10

fusion2004:
Ah, you're correct, I'm on 2.0 until our app finishes upgrade from rails 7.1 to 7.2. Okay, thanks!

PR #903, opened 2025-05-07
Add support for Date/Time attribute values

### LINKED ISSUE #883: Support `Date` / `Time` attribute values

joeldrapper:
Closes #883

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -481,6 +481,8 @@ class Phlex::SGML
 				v.name.tr("_", "-").gsub('"', "&quot;")
 			when Integer, Float
 				v.to_s
+			when Date, Time
+				v.iso8601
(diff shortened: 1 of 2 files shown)


lib/phlex/sgml.rb
```diff
@@ -481,6 +481,10 @@ def json_escape(string)
 				v.name.tr("_", "-").gsub('"', "&quot;")
 			when Integer, Float
 				v.to_s
+			when Date
+				v.iso8601
+			when Time
+				v.respond_to?(:iso8601) ? v.iso8601 : v.strftime("%Y-%m-%dT%H:%M:%S%:z")
```
davetron5000, line 487: Shouldn't this be `:%z`?

davetron5000: Huh, TIL

Issue #882, opened 2025-03-19
Support comma-separated lists


joeldrapper:
Some specific attributes on specific elements should be comma-separated when an Array is passed.

PR #904, opened 2025-05-07
Comma-separated tokens

joeldrapper:
WIP implementation of #882

### DIFF
--- a/lib/phlex/sgml/elements.rb
+++ b/lib/phlex/sgml/elements.rb
@@ -20,7 +20,7 @@ module Phlex::SGML::Elements
 				rel_attribute = attributes[:rel] || attributes["rel"]
 				as_attribute = attributes[:as] || attributes["as"]
 
-				if (:preload == rel_attribute || "preload" == rel_attribute) && (:image == as_attribute || "image" == as_attribute)
+				if ("preload" == rel_attribute || :preload == rel_attribute) && ("image" == as_attribute || :image == as_attribute)
 					attributes[:imagesrcset] = __nested_tokens__(imagesrcset_attribute, ", ")
 				end
 			end
@@ -29,7 +29,7 @@ module Phlex::SGML::Elements
 			if Array === (accept_attribute = attributes[:accept])
 				type_attribute = attributes[:type] || attributes["type"]
 
-				if :file == type_attribute || "file" == type_attribute
+				if "file" == type_attribute || :file == type_attribute
 					attributes[:accept] = __nested_tokens__(accept_attribute, ", ")
 				end
 			end
--- a/lib/phlex/sgml/elements.rb
+++ b/lib/phlex/sgml/elements.rb
@@ -34,15 +34,6 @@ module Phlex::SGML::Elements
 				end
 			end
 		RUBY
-		meta: <<~RUBY,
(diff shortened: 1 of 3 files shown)


lib/phlex/sgml/elements.rb
```diff
@@ -1,6 +1,39 @@
 # frozen_string_literal: true
 
 module Phlex::SGML::Elements
+	COMMA_SEPARATED_TOKENS = {
+		img: <<~RUBY,
+			if (tokens = attributes[:srcset])
+				attributes[:srcset] = __nested_tokens__(tokens, ", ")
+			end
+		RUBY
```
joeldrapper, line 9: The technique here is to generate specifically optimised code to check and pre-process the comma-separated-lists before caching takes place.

lib/phlex/sgml/elements.rb
```diff
@@ -23,6 +56,7 @@ def #{method_name}(**attributes)
 					if block_given # with content block
 						buffer << "<#{tag}"
 						begin
+							#{COMMA_SEPARATED_TOKENS[method_name]}
```
joeldrapper, line 59: This outputs nothing for most elements and otherwise outputs the code to pre-process one or two values if they exist before caching.

lib/phlex/sgml/elements.rb
```diff
@@ -1,6 +1,39 @@
...
 
 module Phlex::SGML::Elements
+	COMMA_SEPARATED_TOKENS = {
+		img: <<~RUBY,
+			if Array === (tokens = attributes[:srcset])
+				attributes[:srcset] = __nested_tokens__(tokens, ", ")
+			end
+		RUBY
+		link: <<~RUBY,
+			if Array === (media_tokens = attributes[:href])
```
willcosgrove, line 11:
```suggestion
			if Array === (media_tokens = attributes[:media])
```

lib/phlex/sgml/elements.rb
```diff
@@ -1,6 +1,39 @@
...
+			end
+
+			if Array === (sizes_tokens = attributes[:sizes])
+				attributes[:sizes] = __nested_tokens__(sizes_tokens, ", ")
+			end
+		RUBY
+		input: <<~RUBY,
+			input_type = attributes[:type]
+
+			if Array === (tokens = attributes[:href]) && (:file == input_type || "file" == input_type)
```
willcosgrove, line 22:
```suggestion
			if Array === (tokens = attributes[:accept]) && (:file == input_type || "file" == input_type)
```

willcosgrove (approved):
Concept ACK. I like this approach a lot. It's nice that it only adds the overhead to elements that may use it. I left a few typo fixes.

lib/phlex/sgml/elements.rb
joeldrapper: These were sneaky AI "fixes". 😳

willcosgrove:
I was looking back at this today as I was preparing to use it for `srcset` on some images and I'm now wondering if we should do any kind of comma escaping for elements in a comma separated token list. Specifically I'm generating image urls using cloudflare's image transforming service, which has options specified in a comma separated list in the URI. In a `srcset` they would need to be escaped into `%2C`.

PR #916, opened 2025-07-05
Prototype Compiler

joeldrapper:


lib/phlex/compiler/method_compiler.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+
+class Phlex::Compiler::MethodCompiler < Prism::Visitor
+	def initialize(compiler)
+		@compiler = compiler
+		@mutations = []
+		@in_scope = true
```
joeldrapper, line 7: This tracks if we may have gone out of scope via a block to an unknown method. Out of scope, we need to use a conditional to check that we’re in scope at runtime. Example

```ruby
10.times do
  if self == component_instance
    # compiled code here
  else
    # original code here
  end
end
```

lib/phlex/compiler/class_compiler.rb
```diff
@@ -0,0 +1,27 @@
...
+		nil
+	end
+
+	def visit_module_node(node)
+		nil
+	end
+
+	def visit_block_node(node)
+		nil
+	end
```
joeldrapper, line 26: We don’t want to visit other classes, modules or blocks since any method definitions here may be in the wrong scope.

lib/phlex/compiler/class_compiler.rb
```diff
@@ -0,0 +1,27 @@
...
+		@compiler = compiler
+	end
+
+	def compile(node)
+		visit_all(node.child_nodes)
+	end
+
+	def visit_def_node(node)
+		Phlex::Compiler::MethodCompiler.new(@compiler).compile(node)
+	end
```
joeldrapper, line 14: We may need to check the receiver of this  method definition. Not sure if this node would be triggered for something like `def Foo.bar`

lib/phlex/compiler.rb
```diff
@@ -0,0 +1,27 @@
...
+		new(component, path, line, source, tree).compile
+	end
+
+	def initialize(component, path, line, source, tree)
+		@component = component
+		@path = path
+		@line = line
+		@source = source
+		@tree = tree
+	end
```
joeldrapper, line 20: We should freeze this.

lib/phlex/compiler/file_compiler.rb
```diff
@@ -0,0 +1,29 @@
...
+	def initialize(compiler)
+		@compiler = compiler
+	end
+
+	def compile(node)
+		visit(node)
+	end
+
+	def visit_class_node(node)
+		if @compiler.line == node.location.start_line
```
joeldrapper, line 13: It’s possible to define more than one class on a single line, so we should also check the name.

lib/phlex/compiler/method_compiler.rb
```diff
@@ -0,0 +1,109 @@
...
+		visit_all(node.child_nodes)
+
+		method_offset = node.location.start_offset
+		source = @compiler.source.byteslice(method_offset, node.location.length)
+
+		if @mutations.any?
+			@mutations.sort_by! { |(offset, length, replacement)| offset }
+
+			@mutations.reverse_each do |(offset, length, replacement)|
+				source[offset - method_offset, length] = replacement
```
joeldrapper, line 20: I don’t know if this is byte based or character based. Need to confirm.

PR #929, opened 2025-07-08
Add an equivalence test harness for the compiler with a failing test

willcosgrove:


quickdraw/compilation_equivalence.test.rb
```diff
@@ -0,0 +1,16 @@
...
+		load File.expand_path(file, File.dirname(__FILE__))
+
+		class_name = File.basename(file, ".rb").split("_").map(&:capitalize).join
+		component = Object.const_get(class_name)
+
+		before = component.new.call
+		Phlex::Compiler.compile(component)
+		after = component.new.call
+
+		assert_equal after, before
```
joeldrapper, line 14: Should we use `assert_equal_html` here?

joeldrapper (approved):
Awesome!

quickdraw/compilation_equivalence.test.rb
willcosgrove: Yes we should!

willcosgrove: CleanShot 2025-07-08 at 13 20 24@2x
That's better!

joeldrapper: The reason this isn’t working is we haven’t got support for void elements yet.

PR #932, opened 2025-07-08
Generate source maps

### LINKED ISSUE #919: Generate source maps
We should be able to create source maps from the compiled source back to the original source code.

joeldrapper:
WIP support for source maps. This doesn't include a hook to apply the source maps to exceptions but that should be relatively easy to do later on.

Closes #919

### DIFF
--- a/lib/phlex/compiler/formatter.rb
+++ b/lib/phlex/compiler/formatter.rb
@@ -36,7 +36,7 @@ module Phlex::Compiler
 		def format(node)
 			@buffer.clear
 			visit(node)
-			@buffer.join
+			[@buffer.join, @source_map]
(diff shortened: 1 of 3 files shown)


lib/phlex/compiler/formatter.rb
```diff
@@ -36,7 +36,7 @@ def visit_array(nodes)
 		def format(node)
 			@buffer.clear
 			visit(node)
-			@buffer.join
+			[@buffer.join, @source_map]
```
joeldrapper, line 39: Probably should return a more sensible object at some point.

willcosgrove (approved):
Nice! This makes sense

PR #935, opened 2025-07-09
Support selective rendering in compiled components

joeldrapper:
Input

```ruby
def view_template
	h1 { "Hello" }
	fragment :name do
		h2 { "World" }
	end
end
```

Output

```ruby
def view_template
	__phlex_state__ = @_state
	__phlex_buffer__ = __phlex_state__.buffer
	__phlex_me__ = self
	__phlex_should_render__ = __phlex_state__.should_render?; nil
	__phlex_buffer__ << "<h1>Hello</h1>" if __phlex_should_render__; nil;
	fragment :name do
		__phlex_original_should_render__ = __phlex_should_render__
		__phlex_should_render__ = __phlex_state__.should_render?;
			__phlex_buffer__ << "<h2>World</h2>" if __phlex_should_render__; nil;
		__phlex_should_render__ = __phlex_original_should_render__
	end
end
```

### DIFF
--- a/lib/phlex/compiler/method_compiler.rb
+++ b/lib/phlex/compiler/method_compiler.rb
@@ -188,7 +188,7 @@ module Phlex::Compiler
 				body: [
 					statement("__phlex_original_should_render__ = __phlex_should_render__"),
 					statement("__phlex_should_render__ = __phlex_state__.should_render?;"),
-					compile_block_body_node(node.body),
+					visit(node.body),
 					statement("__phlex_should_render__ = __phlex_original_should_render__"),
 				]
 			)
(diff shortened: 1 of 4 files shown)


lib/phlex/compiler/method_compiler.rb
```diff
@@ -165,6 +177,23 @@ def compile_plain_helper(node)
...
+			)
+		end
+
+		def compile_fragment_helper_block(node)
+			node.copy(
+				body: [
+					statement("__phlex_original_should_render__ = __phlex_should_render__"),
+					statement("__phlex_should_render__ = __phlex_state__.should_render?;"),
+					compile_block_body_node(node.body),
+					statement("__phlex_should_render__ = __phlex_original_should_render__"),
```
joeldrapper, line 192: Because the local is first created at the top of the method, we need to put it back after changing it.

willcosgrove (approved):

PR #936, opened 2025-07-09
Add `comment` and `raw` to compiler

willcosgrove:
I didn't think there was really any optimization to be done to `raw` but it needs to split the buffer when it's encountered.


lib/phlex/compiler/method_compiler.rb
```diff
@@ -194,6 +198,19 @@ def compile_fragment_helper_block(node)
 			)
 		end
 
+		def compile_comment_helper(node)
+			[
+				buffer("<!-- "),
+				visit_phlex_block(node.block),
```
joeldrapper, line 204: We may be able to just visit the block body here so it doesn't split the buffer.

PR #937, opened 2025-07-09
Add compiler awareness for &block arguments

willcosgrove:


lib/phlex/compiler/method_compiler.rb
```diff
@@ -97,7 +97,9 @@ def visit_phlex_attributes(node)
 		end
 
 		def visit_phlex_block(node)
-			if output_block?(node)
+			if Prism::BlockArgumentNode === node
+				[statement("__yield_content__("), node, push(")")]
+			elsif output_block?(node)
```
willcosgrove, line 102: I added this at the top of the `if` branches because both `output_block?` and `content_block?` assume they're *not* dealing with a `Prism::BlockArgumentNode`, and will blow up if given one.

PR #948, opened 2025-07-21
Use Refract

joeldrapper:


lib/phlex/compiler.rb
```diff
@@ -1,15 +1,28 @@
 # frozen_string_literal: true
 
 require "prism"
+require "refract"
 
 module Phlex::Compiler
 	Error = Class.new(StandardError)
 
+	class BufferPush
```
willcosgrove, line 9: Is this a vestige of a previous attempt?

PR #986, opened 2026-07-14
Fix Ruby 4.0 warning in default before_template and after_template

### LINKED ISSUE #985: Ruby 4.0 warning: block passed to before_template/after_template "may be ignored"
### Summary

On Ruby 4.0, rendering any component emits two verbose warnings because `before_template` / `after_template` are *called* with a block but their definitions don't declare a block parameter:

```
phlex-2.4.1/lib/phlex/sgml.rb:67: warning: the block passed to 'Phlex::SGML#before_template' defined at phlex-2.4.1/lib/phlex/sgml.rb:327 may be ignored
phlex-2.4.1/lib/phlex/sgml.rb:83: warning: the block passed to 'Phlex::SGML#after_template' defined at phlex-2.4.1/lib/phlex/sgml.rb:331 may be ignored
```

### Environment

- phlex 2.4.1 (also present on `main`)
- Ruby 4.0.5 (`+PRISM [arm64-darwin23]`)

### Root cause

Call sites pass the content block (`lib/phlex/sgml.rb:67` and `:83`):

```ruby
before_template(&block)
# ...
after_template(&block)
```

but the default hook definitions take no block (`lib/phlex/sgml.rb:327` and `:331`):

```ruby
private def before_template
  nil
end

private def after_template
  nil
end
```

Ruby 4.0 warns when a block is forwarded to a method whose definition declares no block parameter. The same shape is on `main` (call sites ~57–58, defs ~284–289), so it isn't fixed in an unreleased version.

### Reproduction

Render any component with Ruby warnings enabled, e.g. `RUBYOPT=-w bundle exec rspec` (or `ruby -w`).

### Suggested fix

Declare the block parameter on both hooks so the warning clears and the content block stays available to overrides:

```ruby
private def before_template(&)
  nil
end

private def after_template(&)
  nil
end
```

Alternatively, drop `&block` from the two call sites if these hooks aren't meant to receive the content block. The `(&)` approach preserves the apparent intent of forwarding the block to overrides.

ErnaneJ:
Fixes #985

`SGML#call` passes the caller's block to `before_template` and `after_template` via `&block`, but the default no-op implementations had no block parameter. Ruby 4.0 warns about this:
```
warning: the block passed to 'before_template' may be ignored
warning: the block passed to 'after_template' may be ignored
```
I added `(&)` to both signatures so they accept the block without binding it to a variable. Subclasses that override these methods won't be affected.

### DIFF
--- a/lib/phlex/sgml.rb
+++ b/lib/phlex/sgml.rb
@@ -326,11 +326,11 @@ class Phlex::SGML
 		nil
 	end
 
-	private def before_template
+	private def before_template(&)
 		nil
 	end
 
-	private def after_template
+	private def after_template(&)
 		nil
 	end
 
--- a/quickdraw/sgml/callbacks.test.rb
+++ b/quickdraw/sgml/callbacks.test.rb
@@ -30,4 +30,9 @@ class CallbacksTest < Quickdraw::Test
 	test "callbacks are called in the correct order" do
 		assert_equal Example.call, "<i>1</i><i>2</i><i>3</i><i>4</i><i>5</i><i>6</i><i>7</i>"
 	end
+
+	test "default before_template and after_template accept a block" do
+		klass = Class.new(Phlex::HTML) { def view_template = span { "hello" } }
+		assert_equal klass.call { "ignored" }, "<span>hello</span>"
+	end
 end


joeldrapper:
Thanks

PR #987, opened 2026-07-27
Update documentation with forum

solojavier:
Adding links to the forum. Happy to make any updates or adjustments as needed.

It might be good to disable GitHub discussions to avoid any confusion too.


joeldrapper:
I think disabling discussions would remove existing discussions from the internet. Not 100% sure though.

solojavier:
Thanks! I didn't realized you had a lot of discussions there already. If you're open to the idea I could check if it's possible to migrate them.

